PR #12, opened 2021-12-21
Produce error message if file not found.

kant2002:


kant2002:
Pobably this is become a question. It's quite annoying to see stack trace for handled cases. Are this is nescessary?

ForNeVeR:
Generally, I think that the compiler should only show its stack traces in case of ICE (and internal compiler error), which is _always_ considered a bug.

So, we should handle every known error possible, yes.

The changes are good. Could you please rebase/merge with `main`?

ForNeVeR (approved):

ForNeVeR:
Thanks!

Issue #10, opened 2021-12-21
Support different target frameworks


ForNeVeR:
As of now, Cesium-generated binaries are detected as ".NET Framework" ones with a decompiler (dotPeek in my case).

This doesn't matter much, but we should take a look into this and allow the user to choose the target framework they desire (be it .NET Standard, any version of .NET Core or .NET 5+).

Probably, this is being detected based on our `mscorlib` reference, which should be controlled.

PR #15, opened 2021-12-21
Supported different target frameworks

kekyo:
I tried implementing #10 .

```csharp
    public enum TargetFrameworks
    {
        mscorlib,
        SystemRuntime,
        netstandard,
    }

    public record TargetRuntimeIdentifier(
        TargetFrameworks targetFramework,
        Version version);

    public static AssemblyDefinition GenerateAssembly(
        TranslationUnit translationUnit,
        AssemblyNameDefinition name,
        ModuleKind kind,
        TargetRuntimeIdentifier? targetRuntime = default) { ... }
```

`TargetRuntimeIdentifier` is producing our target framework identity to cecil.
Defaulted target framework (`default`) is netstandard2.1 .

Is issue right for meaning it?


ForNeVeR:
Hey, thank you! This is definitely part of #10, but that's not all.

I've found that the decompiler (at least dotPeek I use) relies on `TargetFrameworkAttribute` being set for the target assembly.

I'll make it work, too.

PR #17, opened 2021-12-21
Add test for C identifier rules

kant2002:
My apologies for creative tests which make your code suffer.

### DIFF
--- a/Cesium.Parser.Tests/LexerTests.cs
+++ b/Cesium.Parser.Tests/LexerTests.cs
@@ -33,4 +33,67 @@ public class LexerTests
             "CloseBrace: }"
         }, tokens);
     }
+
+    [Fact]
+    public void ParametersTest()
+    {
+        const string source = "int main(int argc, char  *argv [ ]) {}";
+        var tokens = GetTokens(source).Select(t => $"{t.Kind}: {t.Text}");
+        Assert.Equal(new[]
+        {
+            "KeywordInt: int",
+            "Identifier: main",
+            "OpenParen: (",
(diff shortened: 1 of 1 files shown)


ForNeVeR:
This is a right direction, but at least some of the failures are caused by the simplicity of [the lexer we've taken from the Yoakke library](https://github.com/LanguageDev/Yoakke/blob/fcc4b7eb0c5180cbe4bad34b4f0c29d26fafb32f/Sources/Core/Yoakke.C.Syntax/CLexer.cs), which doesn't support Unicode, yet.

We may either improve Yoakke lexer (I think that the author would approve that), or just make our own.

I guess I could merge these (in disabled state) and open an issue for someone to pick up (or I can work on this myself later).

kant2002:
After looking at what Roslyn does, I think all "lexers" which are trying to emulate YACC is doomed to fail for compilers.
That idea was only confirmed by https://www.youtube.com/watch?v=wSdV1M7n4gQ
At least 2 languages C# and TypeScript share this philosophy. (What a conicidence). Interestig to hear how other languages look at this problem.

So in short, I definitely propose create own parsers, and our own tokens (also, Yoakke missing whitespaces during lexing phase) but for now that's overkill.

ForNeVeR:
To clarify: yes, we _could_ migrate to a more advanced parser in the future (this project is not an example of using a parser generator). And, yeah, I would like Cesium to fully support Unicode and every strange glyph in a sensible manner.

ForNeVeR (approved):

PR #26, opened 2021-12-22
Explicitly ask in tests for Unicode characters support

### LINKED ISSUE #18: Unicode support in lexer
Currently, we use [a lexer we've taken from the Yoakke library](https://github.com/LanguageDev/Yoakke/blob/fcc4b7eb0c5180cbe4bad34b4f0c29d26fafb32f/Sources/Core/Yoakke.C.Syntax/CLexer.cs), which works really well so far.

One downside is that it doesn't support Unicode (probably because the C standard doesn't mandate its support).

I think that Cesium should fully support Unicode, so we should improve our lexer.

See commented-out tests in https://github.com/ForNeVeR/Cesium/pull/17 which were disabled due to this issue.

kant2002:
Closes #18

### DIFF
--- a/Cesium.Parser.Tests/LexerTests.cs
+++ b/Cesium.Parser.Tests/LexerTests.cs
@@ -85,9 +85,6 @@ public class LexerTests
         yield return new object[] { "9G" };
 
         // This is valid tests which cannot be expressed in Yoakke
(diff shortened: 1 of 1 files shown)


ForNeVeR (approved):
Thanks!

PR #27, opened 2021-12-22
Add reference to standard

kant2002:


ForNeVeR:
You were referring to C11, while we're targeting C17 here. I've updated the link with something which the standard committee consider to be a "similar draft" to the C17 standard.

ForNeVeR:
Thanks for help!

PR #31, opened 2021-12-29
Add ability to suppress compiler logo

kant2002:
I do have plans to use that in the E2E tests, to reduce verboseness of comparison


ForNeVeR (approved):
Thanks!

PR #32, opened 2021-12-29
Add basic VSCode configuration

kant2002:
I'm not of a fan of using VS Code over VS, but on my Gentoo laptop at home this is best option IMO.


ForNeVeR (approved):
I don't use VSCode for .NET development and thus I cannot validate these changes, but I see no harm in merging them. Thanks!

PR #33, opened 2021-12-29
Add basic functional tests

kant2002:
- Compare results of CL.exe compilation with cesium.

Not sure if this test framework will fly or not, and it is not working on Linux yet. 
Based on my understanding how generators should be tested this is best long-term approach.


ForNeVeR:
I'm okay with the idea in general, but will make some changes in the script and wire it up with the CI.

kant2002:
I would like to hear more. I work from Windows and Linux so this is at least half of the work. Also I would like tot use this for testing basic runtime library.

.gitignore
```diff
@@ -4,7 +4,4 @@
 bin/
 obj/
 
-*.dll
```
kant2002, line 7: Can you leave this? If you run commands from readme it pollutes `git status`

ForNeVeR: Yeah, sure.

ForNeVeR (approved):
Thanks for help! Now this will work in CI.

PR #34, opened 2021-12-29
Add basic runtime library

### LINKED ISSUE #23: Basic runtime library
For now, I think Cesium should have a runtime library in a separate assembly (written in C#). This should define nearly all of the C standard library facilities.

I have considered other approaches (write it in C with CLI interop and compile separately with Cesium, write all the code in headers), and consider the current idea the best for the development speed of this part.

This task will be considered closed after we add a runtime library with a couple of standard functions; no fancy stuff is required.

See also #22 (about a way to call the runtime library functions).

kant2002:
- puts
- strlen

Closes #23

### DIFF
--- a/Cesium.IntegrationTests/stdio_test.c
+++ b/Cesium.IntegrationTests/stdio_test.c
@@ -1,0 +1,7 @@
+#include <stdio.h>
+
+int main()
+{
+    puts("test");
+    return 42;
+}
--- a/Cesium.IntegrationTests/stdio_test.c
+++ b/Cesium.IntegrationTests/stdio_test.c
@@ -1,7 +1,0 @@
-#include <stdio.h>
-
-int main()
-{
-    puts("test");
-    return 42;
-}
--- a/Cesium.Runtime/Cesium.Runtime.csproj
+++ b/Cesium.Runtime/Cesium.Runtime.csproj
@@ -1,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+</Project>
--- a/Cesium.Runtime/Cesium.Runtime.csproj
+++ b/Cesium.Runtime/Cesium.Runtime.csproj
@@ -1,10 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
 </Project>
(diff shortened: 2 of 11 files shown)


kant2002:
Errors seems to be from infra

Cesium.Runtime/Cesium.Runtime.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
ForNeVeR, line 4: What do you think of targeting netstandard2.0 in our runtime library? I intend to support .NET Framework (not all versions, but at least the most recent ones) if it's not too tedious.

Cesium.IntegrationTests/stdio_test.c
```diff
@@ -0,0 +1,7 @@
+#include <stdio.h>
+
+int main()
+{
+    puts("test");
```
ForNeVeR, line 5: Probably this fails to parse due to string literals not being supported, yet.

Cesium.Runtime/Cesium.Runtime.csproj
kant2002: Will change. I think that would be fine.

Cesium.IntegrationTests/stdio_test.c
kant2002: I will play. Maybe I will change test somehow

Cesium.Runtime/Cesium.Runtime.csproj
kant2002: If we insist on having char as UTF-8 char, then for netstandard2.0 I cannot use `Marshal.PtrToStringUTF8` and we have to manually convert string to UTF8.

Cesium.IntegrationTests/stdio_test.c
kant2002: There nothing which I can do. I try following variants with same error in compiler, so I probably have to remove this file from integration tests.
```
//char t[5] = { 't', 'e', 's', 't', '\0', };
char t[5] = { 0x74, 0x65, 0x73, 0x74, 0, };
puts(t);
//puts("test");
```

Cesium.Runtime/Cesium.Runtime.csproj
ForNeVeR: Hm, alright. For the time being, let's live with `net6.0`, then.

Cesium.IntegrationTests/stdio_test.c
ForNeVeR: All of these aren't supported, yet :)

I propose to put this aside for now (a couple of days max; probably a dozen of hours) and give me some time to implement at least `const char* val = "hello world";`

kant2002: I think if I can add some function which does not rely on char, maybe math functions and test these parts of runtime lib. That way we will move forward.

Cesium.Runtime/Cesium.Runtime.csproj
kant2002: I have both `netstandard2.0` version and `net6.0` version. Should I remove `netstandard2.0` ?

ForNeVeR: I see, you've implemented the method yourself.

Let's leave this in the current state, then. `netstandard2.0;net6.0` looks okay to me for now.

Cesium.IntegrationTests/stdio_test.c
ForNeVeR: That would be okay, yeah. Feel free to omit this particular integration test for now.

kant2002:
I have no idea why this is green. This is consistently fails on my Linux (Gentoo) machine.

ForNeVeR (approved):
This is awesome!

PR #42, opened 2021-12-31
Tests include path resolution

kant2002:
Right now includes works a bit strange. I assume that `""` resolution should start from location of main compilation file, not from current directory.

### DIFF
--- a/Cesium.Compiler/Program.cs
+++ b/Cesium.Compiler/Program.cs
@@ -10,13 +10,13 @@ using Mono.Cecil;
 using Yoakke.C.Syntax;
 using Yoakke.Streams;
 
-static Task<string> Preprocess(TextReader reader)
+static Task<string> Preprocess(string compilationFileDirectory, TextReader reader)
 {
     var currentProcessPath = Path.GetDirectoryName(Environment.ProcessPath)
                              ?? throw new Exception("Cannot determine path to the compiler executable.");
 
     var stdLibDirectory = Path.Combine(currentProcessPath, "stdlib");
-    var includeContext = new FileSystemIncludeContext(stdLibDirectory, Environment.CurrentDirectory);
+    var includeContext = new FileSystemIncludeContext(stdLibDirectory, compilationFileDirectory);
     var preprocessorLexer = new CPreprocessorLexer(reader);
     var preprocessor = new CPreprocessor(preprocessorLexer, includeContext);
     return preprocessor.ProcessSource();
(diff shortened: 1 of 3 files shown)


kant2002:
This non-failure quite strange.

```
~/Cesium $ pwsh -c ./Cesium.IntegrationTests/Run-Tests.ps1
....
Unhandled exception. System.IO.FileNotFoundException: Could not find file '~/Cesium/empty_include.h'.
File name: '~/Cesium/empty_include.h'
```

ForNeVeR:
> Right now includes works a bit strange. I assume that `""` resolution should start from location of main compilation file, not from current directory.

Sadly, I wasn't able to find anything about that said in the standard, but I'm okay with what you propose.

Right now it tries to resolve paths from the current directory, yeah.

ForNeVeR:
… and, yes, it is strange that the test doesn't fail :)

kant2002:
I will have to troubleshoot this failing tests, because this is consistently fails locally. I can test on Windows only in hour or so, but if you experience the same issue with tests on this branch on Win, let me know.

ForNeVeR:
The issue should be fixed by #43. Do you want to keep this open to track the fix for include lookup?

kant2002:
> The issue should be fixed by #43. Do you want to keep this open to track the fix for include lookup?

Yes. I initially create this to keep track of the problem and provide fix. But was puzzled that it is green. Now after rebase it is failing again and I would move towards fixing this test.

kant2002:
From https://docs.microsoft.com/en-us/cpp/build/reference/i-additional-include-directories?view=msvc-170

The compiler searches directories in the following order:

1. If the **#include** directive is specified using double-quote form, it first searches local directories. The search begins in the same directory as the file that contains the **#include** directive. If it fails to find the file, it searches next in the directories of the currently opened include files, in the reverse order in which they were opened. The search begins in the directory of the parent include file and continues upward through the directories of any grandparent include files.

2. If the **#include** directive is specified in angle-bracket form, or if the local directory search has failed, it searches directories specified by using the /I option, in the order they're specified on the command line.

3. Directories specified in the **INCLUDE** environment variable.

ForNeVeR:
Not the standard, but sounds reasonable to me. Let's consider this as our target in include resolution.

kant2002:
GCC docs https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html
Cannot find Clang official docs, but here good link https://discourse.mc-stan.org/t/clang-include-path-order/2787/2

All of them lookup quoted header file in the same directory as file currently compiling first. GCC seems to be have most simple search strategy.

kant2002:
Standard says only that quoted headers resolved in implementation specific manner. Since all 3 major compilers have some similarities we should try to follow what's people already doing.

ForNeVeR (approved):
Thanks!

PR #44, opened 2022-01-01
Basic string literal support

### LINKED ISSUE #24: Basic string support
Cesium should be able to compile programs with simple string literals (no `wchar_t` or fancy `u8` prefixes, yet; just the basic stuff).

Simple character strings should be compiled down to pinned byte arrays encoded with UTF-8.

### LINKED ISSUE #25: Const support
C programming language includes a complex `const` mechanism to declare objects as `const`. Cesium should support that.

In the nearest future, we'll probably only get an ability to ignore the `const` modifier (because the standard library declares certain arguments as `const`, and I don't want to drop any pieces from the standard library definitions), but eventually we should get the full `const` analysis.

ForNeVeR:
We do the same trick as Roslyn does for the readonly byte arrays.

1. For each literal length, we emit a special `<ConstantPoolItem{length}>` type of desired size.
2. We emit the `<ConstantPool>` type containing a field for each separate string constant, which is being initialized by a data constant.
3. When necessary, we get pointers to these static fields (which are static and thus are safely addressable via pointers).

Closes #24.
Closes #25.


Cesium.Parser.Tests/ParserTests/FullParserTests.cs
```diff
@@ -67,4 +67,16 @@ int main()
...
 }");
+
+    [Fact]
+    public Task StringLiteral() => DoTest(@"int main()
+{
+    test(""hello world"");
+}");
+
+    [Fact]
+    public Task StringVariable() => DoTest(@"int main()
```
kant2002, line 78: Does current implementation handle case of non-const `char*`?

Cesium.Parser/CParser.cs
```diff
@@ -158,13 +161,19 @@ public partial class CParser
...
-        rest?.Insert(0, typeSpecifier) ?? ImmutableArray.Create((DeclarationSpecifier)typeSpecifier);
+        rest?.Insert(0, typeSpecifier) ?? ImmutableArray.Create<IDeclarationSpecifier>(typeSpecifier);
+
+    [Rule("declaration_specifiers: type_qualifier declaration_specifiers?")]
+    private static DeclarationSpecifiers MakeDeclarationSpecifiers(
+        TypeQualifier typeQualifier,
+        DeclarationSpecifiers? rest) =>
+        rest?.Insert(0, typeQualifier) ?? ImmutableArray.Create<IDeclarationSpecifier>(typeQualifier);
+
+    // TODO: [Rule("declaration_specifiers: function_specifier declaration_specifiers?")]
```
kant2002, line 175: Shouldn't these TODO located near `MakeDeclarationSpecifiers`?

kant2002 (commented):
Has couple questions

Cesium.Parser.Tests/ParserTests/FullParserTests.cs
ForNeVeR: As of now, `const` is read and then almost completely ignored by the compiler.

(the only exception is that we'll throw an error on `const const` declaration; that's it)

Also, I've tried reading the standard yesterday, but failed to figure out if it's even allowed to have a non-`const` string literal. Is it?

Cesium.Parser/CParser.cs
ForNeVeR: The order of these TODOs strictly corresponds to the standard, for easier orientation.

PR #46, opened 2022-01-01
Add negative integer support for number literals

kant2002:


ForNeVeR (changes requested):
This is good. Could you please add a couple of tests? Usually, for such changes, I expect at least one new `FullParserTest` and a new `CodeGenTest`.

ForNeVeR (approved):
Nice!

PR #47, opened 2022-01-01
Let compiler decide list assemblies for CLI import

kant2002:
I put list of assemblies for resolving CLI import in the `TranslationUnitContext`. I need this for basic runtime library support, since resolution should happens from runtime library, and I do not want to use `Assembly.GetReferencesAssemblies()` purely, because then I cannot run this compiler via NativeAOT.

Other benefit, is that way we explicitly control list of assemblies to resolve CLI import and then this can be controller by developer via command line arguments.

### DIFF
--- a/Cesium.CodeGen/Contexts/TranslationUnitContext.cs
+++ b/Cesium.CodeGen/Contexts/TranslationUnitContext.cs
@@ -1,8 +1,9 @@
+using System.Reflection;
 using Mono.Cecil;
 
 namespace Cesium.CodeGen.Contexts;
 
-public record TranslationUnitContext(AssemblyContext AssemblyContext)
+public record TranslationUnitContext(AssemblyContext AssemblyContext, Assembly[] ImportAssemblies)
 {
     public AssemblyDefinition Assembly => AssemblyContext.Assembly;
     public ModuleDefinition Module => AssemblyContext.Module;
--- a/Cesium.CodeGen/Contexts/TranslationUnitContext.cs
+++ b/Cesium.CodeGen/Contexts/TranslationUnitContext.cs
@@ -1,9 +1,8 @@
-using System.Reflection;
 using Mono.Cecil;
 
 namespace Cesium.CodeGen.Contexts;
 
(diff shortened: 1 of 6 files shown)


Cesium.CodeGen/Contexts/TranslationUnitContext.cs
```diff
@@ -1,8 +1,9 @@
+using System.Reflection;
 using Mono.Cecil;
 
 namespace Cesium.CodeGen.Contexts;
 
-public record TranslationUnitContext(AssemblyContext AssemblyContext)
+public record TranslationUnitContext(AssemblyContext AssemblyContext, Assembly[] ImportAssemblies)
```
ForNeVeR, line 6: The idea is good, but could you please move that list up to the `AssemblyContext`?

I am already preparing the architecture for the case when there are multiple translation units in a single assembly.

Cesium.CodeGen/Extensions/TypeSystemEx.cs
```diff
@@ -16,13 +17,30 @@ public static class TypeSystemEx
...
-        var type = runtimeAssembly.GetType(typeName);
-        if (type == null) return null;
-        var method = type.GetMethod(methodName);
+        var method = FindType(context.ImportAssemblies, typeName, methodName);
         if (method == null) return null;
 
         return context.Module.ImportReference(method);
     }
+
+    private static MethodInfo? FindType(Assembly[] assemblies, string typeName, string methodName)
```
ForNeVeR, line 26: _(Minor)_ I suggest removing this (both overloads) into `FindMethod`.

ForNeVeR (changes requested):
Overall: the changes are right, this is what's needed to be done.

I have only a couple of minor complaints.

Cesium.CodeGen/Extensions/TypeSystemEx.cs
kant2002: Not clear what do you mean here.

ForNeVeR (approved):
Thanks!

PR #53, opened 2022-01-07
Standard signatures for main, function argument access

### LINKED ISSUE #6: Support standard signatures for main
Right now, only `int main()` is supported.

According to the standard, we should support `int main(void)` as well as `int main(int argc, char * argv[])`.

Interestingly, it tells nothing of `int main()` with an empty parameter list. Perhaps it's a common extension?

### LINKED ISSUE #20: Milestone: Hello, world!
Cesium should be able to compile a "Hello, world" program. Namely, this one:
```c
#include <stdio.h>
int main(void) 
{
  puts("Hello, world!");
}
```

This is a simple program, yet it has several important features we're currently lacking:
- [x] #21: basic preprocessor requirement (`#include`)
- [x] standard library routine (`<stdio.h>` and `puts`)
  - [x] #22: ability to call to CLI methods
  - [x] #23: implement these methods in the runtime library
- [x] #24: string support
- [ ] #6: standard `main` signature, `int main(void)`

ForNeVeR:
1. Closes #6.
2. Provides ability to access the function arguments (in a read-only mode, yet).
3. Closes #20.


Cesium.Parser/CParser.cs
```diff
@@ -318,7 +315,56 @@ public partial class CParser
...
+    private static IDirectAbstractDeclarator MakeDirectAbstractDeclarator(
+        IToken _,
+        AbstractDeclarator abstractDeclarator,
+        IToken __) => new SimpleDirectAbstractDeclarator(abstractDeclarator);
+
+    // HACK:
+    // The original rule in the C11 standard says the following:
+    //     direct_abstract_declarator: direct_abstract_declarator? '[' type_qualifier_list? assignment_expression? ']'
+    //
+    // But here, it's impossible to apply it as-is due to, seemingly, an issue with the Yoakke parser library
```
ForNeVeR, line 342:
```suggestion
    // But here, it's impossible to apply it as-is due to an issue with the Yoakke parser library
```

Cesium.Parser/CParser.cs
```diff
@@ -318,7 +315,56 @@ public partial class CParser
...
+    // (https://github.com/LanguageDev/Yoakke/issues/121). Thus, I has to split it in two:
```
ForNeVeR, line 343:
```suggestion
    // (https://github.com/LanguageDev/Yoakke/issues/121). Thus, it has been split in two:
```

Cesium.Runtime/RuntimeHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Cesium.Runtime;
+
+[SuppressMessage("ReSharper", "UnusedMember.Global")]
+public static unsafe class RuntimeHelpers
+{
+    public static byte*[] ArgsToArgv(string[] strings)
```
ForNeVeR, line 11:
```suggestion
    public static byte*[] ArgsToArgv(string[] args)
```

PR #84, opened 2022-01-21
Preprocessor: Naive implementation of #error

### LINKED ISSUE #64: Preprocessor #error support
Our preprocessor should support the `#error` directive from the C standard.

Newlifer:
Closes #64.

### DIFF
--- a/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
+++ b/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
@@ -38,11 +38,11 @@ int test()
     [Fact]
     public async Task ErrorMsg()
     {
-        NotSupportedException err = await Assert.ThrowsAsync<NotSupportedException>(async () => await DoTest(@"
-#error ""Error message"" test
+        PreprocessorException err = await Assert.ThrowsAsync<PreprocessorException>(async () => await DoTest(
+@"#error ""Error message"" test
 int test()
-{}
-"));
+{}"
+        ));
         Assert.Equal(@"Error: ""Error message"" test", err.Message);
     }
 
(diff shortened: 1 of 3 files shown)


Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
```diff
@@ -33,4 +33,17 @@ int test()
 int test()
 {
 }", new() { ["foo.h"] = "void foo() {}", ["bar.h"] = "int bar = 0;" });
+
+
+    [Fact]
+    public async Task ErrorMsg()
+    {
+        NotSupportedException err = await Assert.ThrowsAsync<NotSupportedException>(async () => await DoTest(@"
```
ForNeVeR, line 41: I think that it should be not a `NotSupportedException`, but something else. Let's create a new exception class for this? `class PreprocessorException : Exception`.

Cesium.Preprocessor/CPreprocessor.cs
```diff
@@ -145,6 +145,16 @@ IToken<CPreprocessorTokenType> ConsumeNext(params CPreprocessorTokenType[] allow
...
                 return tokens.ToList();
             }
+            case "error":
+            {
+                bool hasRemaining;
+                var errorText = new StringBuilder();
+                while ((hasRemaining = enumerator.MoveNext()) && enumerator.Current is var t and not { Kind: NewLine })
+                {
+                    errorText.Append(t.Text);
+                }
```
ForNeVeR, line 155: Since there shouldn't be any `NewLine` in the `enumerator` (`ReadDirectiveLine` drops it before this place is being called), this may be simplified.

```suggestion
                var errorText = new StringBuilder();
                while (enumerator.MoveNext())
                {
                    errorText.Append(enumerator.Current.Text);
                }

```

ForNeVeR (changes requested):
Thanks! There are some changes I'd like you to apply before merging this.

Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
Newlifer: Done!

Cesium.Preprocessor/CPreprocessor.cs
Newlifer: Nice! Done. =)

ForNeVeR (approved):
Thanks!

PR #105, opened 2022-02-13
Add more test cases for function signatures

kant2002:
Fix codegen for empty functions without any statement. Inside IL they are require to have at least one `ret`  instruction inside method body.


ForNeVeR (approved):
Thanks!

PR #106, opened 2022-02-14
Fix parse of forward declaration without parameters

kant2002:


Cesium.Parser/CParser.cs
```diff
@@ -271,7 +271,7 @@ public partial class CParser
         IDirectDeclarator @base,
         ICToken _,
         IdentifierList? identifierList,
-        ICToken __) => new IdentifierListDirectDeclarator(@base, identifierList);
+        ICToken __) => new IdentifierListDirectDeclarator(@base, identifierList ?? IdentifierList.Empty);
```
ForNeVeR, line 274: The general rule I'm trying to follow in the main AST is that if something's defined as optional by the standard, then it should be `null` in the AST.

Otherwise, corner cases such as empty function parameter list vs a void one become much harder to track.

Intermediate AST (aka "IR") may be less restrictive, but regarding this particular change I'm not sure.

ForNeVeR (changes requested):
I guess I am waiting for the green tests before merging this, but please tell me if you'd like to get some help with this.

kant2002:
Based on what I see in #171 this is completed.

ForNeVeR:
Thanks for help!

PR #111, opened 2022-03-08
Feature/98.bit arithmetic

### LINKED ISSUE #98: Bit arithmetic support
Our current milestone requires a bit of bit arithmetic support, like operators `|`, `^`, `&`, `<<`, `>>`.

They should work on `char`, `unsigned char`.

Griboedoff:
Closes #98.


Cesium.Parser/CParser.cs
```diff
@@ -123,6 +123,10 @@ public partial class CParser
     private static Expression MakePrefixIncrementExpression(ICToken _, Expression target) =>
         new PrefixIncrementExpression(target);
 
+    [Rule("unary_expression: '~' unary_expression")]
+    private static Expression MakeBitwiseNotExpression(ICToken _, Expression target) =>
+        new BitwiseNotExpression(target);
+
```
ForNeVeR, line 129: I'd prefer this to be closer to the AST provided by the standard. Something like this:
```suggestion
    [Rule("unary_expression: unary_operator cast_expression")]
    private static Expression MakeBitwiseNotExpression(ICToken @operator, Expression target) =>
        new UnaryExpression(@operator, target);

    // TODO: unary-operator: one of
    //    & * + - !
    [Rule("unary_operator: '~'")]
    // TODO: [Rule("unary_operator: '!'")]
    private static Expression MakeUnaryOperator(ICToken @operator) => @operator;
```

Also note how the AST from the standard dictates changes in the AST records: instead of `Ast.BitwiseNotExpression`, I'd prefer to see `Ast.UnaryExpression`.

You're free to leave `Ir.Expressions.BitwiseNotExpression` as-is, though. That's okay.

ForNeVeR (changes requested):
These changes are very good, but I've found a minor issue.

ForNeVeR (approved):
Awesome, thanks!

Issue #100, opened 2022-02-13
Basic support for the `for` statement


ForNeVeR:
For now, we've decided to not add variable definition support / scope support to the `for` statement.

If viable, any _declaration_ inside of a `for` block should throw `NotImplementedException`.

PR #119, opened 2022-05-02
Feature/100.for cycle

Griboedoff:


Griboedoff:
also i've found that we need break statement, so it will be my next task #120

ForNeVeR (approved):
Wonderful. Thanks!

Griboedoff:
solves #100

PR #124, opened 2022-05-18
Feature/102.array indexing

### LINKED ISSUE #102: Array indexing
We should support array indexing, like this:
```
int x[100];
x[0] = 10;
```

Depends on:
- [x] #101

Griboedoff:
Closes #102 
Restrictions: 
- only primitive type arrays
- no type conversion — you can't set byte to int[]


ForNeVeR:
I'd like to take a look at this after we merge #123, since this builds on top of that (which is okay).

ForNeVeR (approved):
Thanks!

PR #125, opened 2022-05-26
Github Actions: workflows are refactored | readme badges are added

kolosovpetro:
This pull request contains the following changes to the Github Actions CI:

- Common steps like install .NET SDK, restore, build are moved to the separate action so that code is not repeated
- Decorated actions with various icons so that each CI step looks more happy 😄 
- Added a `workflow_dispatch` event so that actions may be triggered manually (if necessary)
- Workflow files are renamed so that they reflect the contents exactly
- Workflow status badges are added to the README.md


ForNeVeR (approved):
Thanks for your help! One detail I've decided to remove is the build status badge.

It occupies space in our Readme, but it is anyway visible on GitHub page, and isn't required in cases when people read the Readme outside of GitHub.

PR #129, opened 2022-07-14
Pointer member access and address-of operator

gsomix:
Short example:
```c
typedef struct { int x; } foo;

int main(void) 
{ 
  foo y; 
  foo *z; 
  z = &y; 
  z->x = 42; 
  return z->x; 
}
```

TODO:
- [x] Modify parser to support `->` operator
- [x] Modify parser to support `&` operator
- [x] Make codegen for getting addess of lvalues
- [x] Make codegen for pointer member access operator
- [x] Tests, tests, tests
- [x] Samples


ForNeVeR (approved):
Awesome!

gsomix:
Thanks!

PR #131, opened 2022-07-16
Help enum values

Fantoom:
In current implementation available values for enum type arguments do not show in help

```
Cesium.Compiler 0.0.0
Copyright (C) 2022 Cesium.Compiler
    -o, --out       Required.
    --framework     (Default: Net)
    --nologo        Suppress compiler banner message
...
```

changed to

```
Cesium.Compiler 0.0.0
Copyright (C) 2022 Cesium.Compiler
    -o, --out       Required.
    --framework     (Default: Net)  Valid values: Net, NetFramework, NetStandard
    --nologo        Suppress compiler banner message
...

### DIFF
--- a/Cesium.Compiler/Main.cs
+++ b/Cesium.Compiler/Main.cs
@@ -1,7 +1,8 @@
-﻿using System.Reflection;
+using System.Reflection;
 using Cesium.CodeGen;
 using Cesium.Compiler;
 using CommandLine;
+using CommandLine.Text;
 
 var parserResult = new Parser(x => x.HelpWriter = null).ParseArguments<Arguments>(args);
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ForNeVeR** [comment]
I think you've lost the error message when Cesium is unable to parse the command-line arguments.

In `main`:
```console
$ dotnet run --project Cesium.Compiler -- --framework omg
Cesium.Compiler 0.0.0
Copyright (C) 2022 Cesium.Compiler

ERROR(S):
  Option 'framework' is defined with a bad format.
  Required option 'o, out' is missing.

  -o, --out       Required.
[…]
```

In your branch:
```console
$ dotnet run --project Cesium.Compiler -- --framework omg
Cesium.Compiler 0.0.0
Copyright (C) 2022 Cesium.Compiler

  -o, --out       Required.
[…]
```

Could you please take a look and fix it? Thanks!

Cesium.Compiler/Main.cs
```diff
@@ -3,7 +3,9 @@
 using Cesium.Compiler;
 using CommandLine;
 
-return await Parser.Default.ParseArguments<Arguments>(args).MapResult(async args =>
+var parserResult = new Parser(x => x.HelpWriter = null).ParseArguments<Arguments>(args);
```
ForNeVeR, line 6: Why do you set `TextWriter = null` here? I don't think it is required.

ForNeVeR (changes requested):

Cesium.Compiler/Main.cs
Fantoom: It prints 2 times in console if not set to null

ForNeVeR (approved):
Thanks!

PR #139, opened 2022-07-21
(#74) Basic `switch-case` support

### LINKED ISSUE #74: Compiler: basic `switch-case` support
The title is self-explanatory: we should support simple `switch-case` over numeric data. No fancy stuff (like jumping between cases or in the middle of a loop) is required, yet.

gsomix:
Closes #74 

- [x] Add labeled and switch statements to parser
- [ ] Add switch statement to IR
- [ ] Write codegen for switch-case
- [ ] Add samples

### DIFF
--- a/Cesium.Parser/CParser.cs
+++ b/Cesium.Parser/CParser.cs
@@ -635,7 +635,7 @@ public partial class CParser
     // TODO: 6.7.10 Static assertions
 
     // 6.8 Statements and blocks
-    // TODO: [Rule("statement: labeled_statement")]
+    [Rule("statement: labeled_statement")]
     [Rule("statement: compound_statement")]
     [Rule("statement: expression_statement")]
     [Rule("statement: selection_statement")]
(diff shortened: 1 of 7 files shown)


Cesium.Parser/CParser.cs
```diff
@@ -635,15 +635,28 @@ private ParseResult<ParameterDeclaration> CustomParseParameterDeclaration(int of
...
     [Rule("statement: jump_statement")]
     private static Statement MakeStatementIdentity(Statement statement) => statement;
 
-    // TODO: 6.8.1 Labeled statements
+    // 6.8.1 Labeled statements
+    [Rule("labeled_statement: Identifier ':' statement")]
+    private static Statement MakeLabelStatement(IToken identifier, IToken _, Statement block) =>
+        new LabelStatement(identifier.Text, block);
+
+    // TODO: constant or constant-expression? What's difference? gsomix.
```
kant2002, line 651: @gsomix constant is just 5 , constant-expression can be 4+5

kant2002:
After https://github.com/ForNeVeR/Cesium/pull/302 lands we can model `if/for/switch/while` as `goto` statements. Also it slightly interferes with this PR in label parser area. I actually borrow code from this PR for these parts.

ForNeVeR:
Superseded by #316.

PR #141, opened 2022-07-23
GlobalFunctions Type

### LINKED ISSUE #35: CLI interop: allow to put the functions to a named static class
Currently, we declare C functions as module-level global functions in CLI.

It may be difficult to interoperate using some languages, so probably we'll need to add an opt-in or even an opt-out mode which will put everything into a static class (not the `<Module>` type).

Fantoom:
Closes #35.

Added GlobalFunctions Type
Change function emit from Module Type to GlobalFunctions

### DIFF
--- a/Cesium.CodeGen/Contexts/AssemblyContext.cs
+++ b/Cesium.CodeGen/Contexts/AssemblyContext.cs
@@ -15,8 +15,6 @@ public class AssemblyContext
     public ModuleDefinition Module { get; }
     public Assembly[] ImportAssemblies { get; }
     public TypeDefinition GlobalType { get; }
-    public string GlobalTypeFQN { get; }
-    public string Namespace { get; }
 
     internal Dictionary<string, FunctionInfo> Functions { get; } = new();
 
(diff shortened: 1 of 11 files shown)


ForNeVeR:
I'd like the design to be the following.
1. There should be a command-line option to pass the global function type name
2. The class should be a `abstract sealed` in CIL terms (also known as a `static class` in C#)
3. It should be possible to pass a FQN (i.e. `--globalclass My.Namespace.MyGlobalFunctions`
4. If nothing passed, Cesium should behave in the old way (i.e. use the module type)
5. Please add unit tests for old and new behavior

Fantoom:
> 4. If nothing passed, Cesium should behave in the old way (i.e. use the module type)

module type or based on #140?

Fantoom:
- [x] 1.
- [x] 2.
- [x] 3.
- [x] 4.

Fantoom:
Should be ready for review

Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -73,8 +77,35 @@ private AssemblyContext(AssemblyDefinition assembly, ModuleDefinition module, As
...
+        {
+            string typeName;
+            string typeNamespace;
 
+            if (!globalTypeFQN.Contains('.'))
+            {
+                typeName = globalTypeFQN;
+                typeNamespace = @namespace;
+                GlobalTypeFQN = $"{@namespace}.{globalTypeFQN}";
+            }
```
ForNeVeR, line 92: This I don't like. It makes impossible to define a namespace for all other types, and then mark the global type as namespace-less. In particular, I believe that `--namespace SomeNS --globalclass GlobalClass` should lead to the generation of a class with FQN `GlobalClass` (i.e. without a namespace).

This would be more flexible for various user demands.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks a lot!

PR #143, opened 2022-07-24
[experimental] CI: enable Sonar

ForNeVeR:
Closes #144 (supersedes).


scripts/run-sonarqube-dotnet.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+if [ -z "$SONAR_TOKEN" ]; then
+  echo "SONAR_TOKEN is not set"
+  exit 1
+fi
+
+if [ -z "$GITHUB_TOKEN" ]; then
+  echo "$GITHUB_TOKEN is not set"
```
kolosovpetro, line 9: I think execution flow will never go inside this if, so it can be removed

ForNeVeR:
Thanks for help, @kolosovpetro!

Issue #75, opened 2022-01-16
Compiler: global variables


ForNeVeR:
Our current milestone, #61, targets a project which have several global variable declarations.

For example, [this one](https://github.com/cr-marcstevens/sha1collisiondetection/blob/b4a7b0b157d08609cbe66ddf919b2aa86c3f16b2/lib/sha1.c#L1860-L1866):
```c
static const unsigned char sha1_padding[64] =
{
	0x80, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
};
```

Cesium should support them.

- [ ] #346
- [ ] See also (and eliminate!) `TODO[#75]` in the code.
Depends on:
- [x] #335

ForNeVeR:
Some tests to consider:
```csharp
    [Fact]
    public Task GlobalVariableOfNamedType() => DoTest(@"typedef int foo;
foo x;");

    [Fact]
    public Task GlobalStructVariable() => DoTest(@"typedef struct { int x; } foo;
foo x;");
```

Issue #92, opened 2022-02-06
Use optimized CIL instructions where possible


ForNeVeR:
CIL includes the instructions like `Ldc_I4_0`, `Ldarg_0` with baked-in indices of various entities.

Cesium should use these instructions where appropriate.

See `TODO[#92]` in the code to get to some of the places where the optimization is required.

PR #146, opened 2022-07-24
Global variable

Fantoom:
Part of #75.


Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -101,6 +107,41 @@ public AssemblyDefinition VerifyAndGetAssembly()
...
+        var context = new TranslationUnitContext(this);
+        var scope = new FunctionScope(context, GlobalTypeStaticCtor);
+        var lastOpcode = GlobalTypeStaticCtor.Body.Instructions.Last();
+        if (lastOpcode.OpCode == OpCodes.Ret)
+        {
+            GlobalTypeStaticCtor.Body.Instructions.Remove(lastOpcode);
+            lval.EmitSetValue(scope, value);
+            GlobalTypeStaticCtor.Body.Instructions.Add(lastOpcode);
+        }
+        else
```
kolosovpetro, line 143: This `else` statement could be omitted calling `return` as last instruction inside `if`.

Also I recommend to watch the video about if-else refactors: https://youtu.be/_ougvb8mT7k

Fantoom: Why?

Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -101,6 +107,41 @@ public AssemblyDefinition VerifyAndGetAssembly()
...
+        return method;
+    }
+
+    internal void AddFieldInitialization(FieldDefinition field, Ir.Expressions.IExpression value)
+    {
+        var lval = new Ir.Expressions.LValues.LValueGlobalVariable(field);
+        var context = new TranslationUnitContext(this);
+        var scope = new FunctionScope(context, GlobalTypeStaticCtor);
+        var lastOpcode = GlobalTypeStaticCtor.Body.Instructions.Last();
+        if (lastOpcode.OpCode == OpCodes.Ret)
```
ForNeVeR, line 137: This code is very strange. How are you even getting this `Ret`? I can't see who puts it there.

Cesium.CodeGen/Ir/Expressions/IdentifierExpression.cs
```diff
@@ -33,10 +33,12 @@ public ILValue Resolve(IDeclarationScope scope)
     {
         scope.Variables.TryGetValue(Identifier, out var var);
         var par = scope.GetParameter(Identifier);
-
+        var global = scope.Context.GlobalType.Fields.FirstOrDefault(x => x.Name == Identifier);
```
ForNeVeR, line 36: Does C allow global shadowing by a function parameter or a local variable?

If not, then you should also increase the `switch` here, and check if there's only one variable available (in the same manner it has been done for `(var, par)`.

Cesium.CodeGen/Ir/Expressions/LValues/LValueGlobalVariable.cs
```diff
@@ -0,0 +1,37 @@
...
+        private readonly FieldDefinition _definition;
+
+        public LValueGlobalVariable(FieldDefinition definition)
+        {
+            _definition = definition;
+        }
+
+        public void EmitGetValue(IDeclarationScope scope)
+        {
+            // TODO[#92]: Special instructions to emit Ldloc_0 etc.
```
ForNeVeR, line 19: Since #92 is already done in `main`, you should perform the same thing here.

Cesium.CodeGen/Ir/TopLevel/TopLevelDeclaration.cs
```diff
@@ -65,18 +65,34 @@ public void EmitTo(TranslationUnitContext context)
...
+            var field = EmitGlobalVariable(context, identifier, type);
             if (initializer != null)
             {
-                throw new NotImplementedException(
-                    $"Declaration {declaration} with initializer {initializer} not supported, yet.");
-                // TODO[#75]: Don't forget to lower the initializer.
+                context.AssemblyContext.AddFieldInitialization(field, initializer);
             }
-
+            continue;
```
ForNeVeR, line 73: This is odd. Why is there `continue` followed by an exception?

Cesium.CodeGen/Ir/TopLevel/TopLevelDeclaration.cs
```diff
@@ -65,18 +65,34 @@ public void EmitTo(TranslationUnitContext context)
                 continue;
             }
 
-            // TODO[#75]: Generate a global variable of type {type, isConst}.
+            var field = EmitGlobalVariable(context, identifier, type);
             if (initializer != null)
             {
-                throw new NotImplementedException(
-                    $"Declaration {declaration} with initializer {initializer} not supported, yet.");
-                // TODO[#75]: Don't forget to lower the initializer.
```
ForNeVeR, line 73: Also, it seems you forgot to lower the initializer?

Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -101,6 +107,41 @@ public AssemblyDefinition VerifyAndGetAssembly()
...
+        MethodDefinition method = new MethodDefinition(".cctor", methodAttributes, Module.TypeSystem.Void);
+        method.Body.Instructions.Add(Instruction.Create(OpCodes.Ret));
+        return method;
+    }
+
+    internal void AddFieldInitialization(FieldDefinition field, Ir.Expressions.IExpression value)
+    {
+        var lval = new Ir.Expressions.LValues.LValueGlobalVariable(field);
+        var context = new TranslationUnitContext(this);
+        var scope = new FunctionScope(context, GlobalTypeStaticCtor);
```
ForNeVeR, line 135: Nope, you shouldn't create these objects here.

Instead of this, the caller should pass a `TranslationUnitContext`, and it should implement `IDeclarationScope` (so you'll be able to pass it to `EmitSetValue`).

Usage of `LValueGlobalVariable` is odd but okay.

ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/AssemblyContext.cs
Fantoom: line 127

ForNeVeR: Then remove it :)

Cesium.CodeGen/Ir/Expressions/IdentifierExpression.cs
Fantoom: Yes, see 6.2.1.4 in  [C17 Language Standard Draft](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2310.pdf)

ForNeVeR (approved):
Thanks for help!

PR #147, opened 2022-07-25
Using optimized CIL instructions (ldc_i4_0 instead of ldc_i4, etc).

### LINKED ISSUE #92: Use optimized CIL instructions where possible

impworks:
Fixes #92.


ForNeVeR (approved):
Awesome, thanks!

Issue #142, opened 2022-07-23
NullReferenceException on an empty method body


ForNeVeR:
For example, compiler will crash when compiling this program:
```c
int main()
{
    int unused;
}
```

Here's a quick test checking that:
```csharp
    [Fact]
    public Task FunctionPtrTest1() => DoTest(@"int main()
{
    int unused;
}");
```

And a stacktrace:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Cecil.Cil.CodeWriter.ComputeHeader()
   at Mono.Cecil.Cil.CodeWriter.WriteResolvedMethodBody(MethodDefinition method)
   at Mono.Cecil.Cil.CodeWriter.WriteMethodBody(MethodDefinition method)
   at Mono.Cecil.MetadataBuilder.AddMethod(MethodDefinition method)
   at Mono.Cecil.MetadataBuilder.AddMethods(TypeDefinition type)
   at Mono.Cecil.MetadataBuilder.AddType(TypeDefinition type)
   at Mono.Cecil.MetadataBuilder.AddTypes()
   at Mono.Cecil.MetadataBuilder.BuildTypes()
   at Mono.Cecil.MetadataBuilder.BuildModule()
   at Mono.Cecil.MetadataBuilder.BuildMetadata()
   at Mono.Cecil.ModuleWriter.BuildMetadata(ModuleDefinition module, MetadataBuilder metadata)
   at Mono.Cecil.ModuleWriter.Write(ModuleDefinition module, Disposable`1 stream, WriterParameters parameters)
   at Mono.Cecil.ModuleWriter.WriteModule(ModuleDefinition module, Disposable`1 stream, WriterParameters parameters)
   at Mono.Cecil.ModuleDefinition.Write(Stream stream, WriterParameters parameters)
   at Mono.Cecil.AssemblyDefinition.Write(Stream stream, WriterParameters parameters)
   at Mono.Cecil.AssemblyDefinition.Write(Stream stream)
   at Cesium.CodeGen.Tests.CodeGenTestBase.EmitAssembly(AssemblyContext context) in C:\Projects\Cesium\Cesium.CodeGen.Tests\CodeGenTestBase.cs:line 60
   at Cesium.CodeGen.Tests.CodeGenTestBase.GenerateAssembly(TargetRuntimeDescriptor runtime, String[] sources) in C:\Projects\Cesium\Cesium.CodeGen.Tests\CodeGenTestBase.cs:line 19
[...]

PR #149, opened 2022-07-25
Empty method body

Fantoom:
#142

### DIFF
--- a/Cesium.CodeGen/Extensions/TypeDefinitionEx.cs
+++ b/Cesium.CodeGen/Extensions/TypeDefinitionEx.cs
@@ -18,7 +18,7 @@ internal static class TypeDefinitionEx
             MethodAttributes.Public | MethodAttributes.Static,
             returnType);
 
-        var _ = method.Body.Instructions; // Initalize instructions with empty collection to avoid NRE
+        var _ = method.Body.Instructions; // Initialize instructions with empty collection to avoid NRE
         AddParameters(context, method, parameters);
         type.Methods.Add(method);
         return method;
--- a/Cesium.CodeGen/Extensions/TypeDefinitionEx.cs
+++ b/Cesium.CodeGen/Extensions/TypeDefinitionEx.cs
@@ -17,6 +17,8 @@ internal static class TypeDefinitionEx
             name,
             MethodAttributes.Public | MethodAttributes.Static,
             returnType);
+
+        var _ = method.Body.Instructions; // Initalize instructions with empty collection to avoid NRE
         AddParameters(context, method, parameters);
         type.Methods.Add(method);
         return method;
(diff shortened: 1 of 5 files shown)


Cesium.CodeGen/Ir/BlockItems/CompoundStatement.cs
```diff
@@ -17,7 +17,9 @@ public CompoundStatement(Ast.CompoundStatement statement)
     {
     }
 
-    public bool HasDefiniteReturn => _items.Count > 0; // TODO[#90]: Better definite return analysis.
+    public bool HasDefiniteReturn =>
+        _items.Count > 0
+        && !_items.All(x =>!(x as DeclarationBlockItem)?.IsScopedIdentifierWithInitalizer?? false); // TODO[#90]: Better definite return analysis.
```
kolosovpetro, line 22: Please, refactor messy predicate on the line 22. At first glance, it is seen that you have double negation which does not make any sense. It is clearly that you can simplify double negation of the part `!_items.All(x =>!(x as DeclarationBlockItem)?.IsScopedIdentifierWithInitalizer?? false)`

Next, consider the following possibilities to make predicate inside extension method `All` looking better:

- Replace `as` with `is` using predicate as follows: `x => x is DeclarationBlockItem decl && !decl.IsScopedIdentifierWithInitalizer`
- Using pattern matching `x => x is DeclarationBlockItem { IsScopedIdentifierWithInitalizer: false }`
- Defining method that compatible with the signature `Func<BlockItem, bool>` and passing it directly inside method `All`
- Combining all previous suggestions

Also, you have a typo: `IsScopedIdentifierWithInitalizer` -> `IsScopedIdentifierWithInitializer`

Cesium.CodeGen/Extensions/TypeDefinitionEx.cs
```diff
@@ -17,6 +17,8 @@ internal static class TypeDefinitionEx
             name,
             MethodAttributes.Public | MethodAttributes.Static,
             returnType);
+
+        var _ = method.Body.Instructions; // Initalize instructions with empty collection to avoid NRE
```
kolosovpetro, line 21: Why do we need this discard local variable?

Please clarify what does it mean the comment `// Initalize instructions with empty collection to avoid NRE` ?

Again typo `Initalize` -> `Initialize`

Cesium.CodeGen/Ir/BlockItems/CompoundStatement.cs
DedSec256, line 22:
```suggestion
        && _items.Any(x =>x is DeclarationBlockItem item ? item.IsScopedIdentifierWithInitializer : true); // TODO[#90]: Better definite return analysis.
```

Fantoom:
Closing this pr in favour of #148

ForNeVeR:
Thanks for your contribution, anyway. You seem to have started a wave of interest towards the project.

PR #150, opened 2022-07-25
Remove try finally

### LINKED ISSUE #128: Get rid of try-finally in the synthetic entrypoint
For whatever reason, I've decided that `fixed (byte** argVPtr = argVCopy)` should compile into `try` + `finally`: https://github.com/ForNeVeR/Cesium/blob/f1cd5b1bfbf2b8ff97fae9927b7461f418f49258/Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs#L204-L221

This is not really the case, as correctly highlighted by @EgorBo; [see this SharpLab link](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgZgAIBXAOwGcBDAMxjpLoGE6Abyp1RdEWMKkkPFHQBiECAAowACzZQAVHQAOASkEBfCaNM96hOQFllUgAx1NUfUPNi6HbAhgATOqoa2np0ALxOUC7uYsKUHvFiiioGANzRoiZxYplGQA):
```
using System;
public unsafe class C {
    
    static void Foo(char* p){}
    
    public void M(string arr) {
        fixed (char* p = arr)
        {
            Foo(p);
        }
    }
}
```

This compiles without any `try`/`finally`.

So, as Rosly doesn't care, Cesium shouldn't emit this `try` + `finally` block in the synthetic entrypoint, either.

Fantoom:
Closes #128

**TODO[F]:**
- [x] ~~add an integration test for the synthetic entrypoint~~: turns out it already exists

### DIFF
--- a/Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
+++ b/Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
@@ -209,21 +209,16 @@ internal class FunctionDefinition : ITopLevelNode
                 instructions.Add(Instruction.Create(OpCodes.Stloc_S, exitCode));
                 instructions.Add(Instruction.Create(OpCodes.Leave_S, atExitLdLocExitCode));
             }
-            // finally: unpin
+            //unpin
             {
                 instructions.Add(Instruction.Create(OpCodes.Ldnull));
                 instructions.Add(Instruction.Create(OpCodes.Stloc_3)); // 3 = argVPinned.Index
-                instructions.Add(Instruction.Create(OpCodes.Endfinally));
             }
-        }
-        // finally
-        {
+
             // Cesium.Runtime.RuntimeHelpers.FreeArgv(argV);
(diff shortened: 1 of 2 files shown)


Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
```diff
@@ -185,12 +185,11 @@ private bool ValidateMainParameters()
         instructions.Add(Instruction.Create(OpCodes.Stloc_1)); // 1 = argV.Index
 
         // try
-        Instruction tryStart, tryEnd;
         Instruction pinStart, pinEnd;
         Instruction unpinStart, unpinEnd;
```
ForNeVeR, line 189: We don't need these variables anymore. Could you please get rid of them, too?

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

Issue #93, opened 2022-02-06
CLI interop: verify signatures


ForNeVeR:
Right now, `__cli_import` doesn't verify the imported method's signature.

It should start doing that.

- [ ] look for `TODO[#93]` in the code to find the places where this change could be applied

PR #152, opened 2022-07-26
Method signature validation in __cli_import.

impworks:
Closes #93.

PR #153, opened 2022-07-26
Verify method signature

### LINKED ISSUE #48: Improve method resolution
Right now method resolution does not look for best suited overload.

This leads to problem when trying to author stdlib. I.e.
```c
__cli_import("System.Math::Abs")
int abs(int n);
```
produce
```
Unhandled exception. System.Reflection.AmbiguousMatchException: Ambiguous match found.
   at System.RuntimeType.GetMethodImplCommon(String name, Int32 genericParameterCount, BindingFlags bindingAttr, Binder binder, CallingConventions callConv, Type[] types, ParameterModifier[] modifiers)
   at System.Type.GetMethod(String name, BindingFlags bindingAttr)
   at System.Type.GetMethod(String name)
   at Cesium.CodeGen.Extensions.TypeSystemEx.FindMethod(Assembly assembly, String typeName, String methodName) in ~/Cesium/Cesium.CodeGen/Extensions/TypeSystemEx.cs:line 43
   at Cesium.CodeGen.Extensions.TypeSystemEx.FindMethod(Assembly[] assemblies, String typeName, String methodName) in ~/Cesium/Cesium.CodeGen/Extensions/TypeSystemEx.cs:line 30
   at Cesium.CodeGen.Extensions.TypeSystemEx.MethodLookup(TranslationUnitContext context, CliImportSpecifier specifier) in ~/Cesium/Cesium.CodeGen/Extensions/TypeSystemEx.cs:line 20
   at Cesium.CodeGen.Generators.Declarations.EmitSymbol(TranslationUnitContext context, SymbolDeclaration symbolDeclaration) in ~/Cesium/Cesium.CodeGen/Generators/Declarations.cs:line 18
   at Cesium.CodeGen.Generators.Assemblies.Generate(TranslationUnit translationUnit, AssemblyNameDefinition name, ModuleKind kind, TargetRuntimeDescriptor targetRuntime, Assembly[] importAssemblies) in ~/Cesium/Cesium.CodeGen/Generators/Assemblies.cs:line 36

```

Fantoom:
Started as a try to solve #93.

Closes #48.


ForNeVeR:
While #93 was fixed by #152, there's still value in this PR, because it also closes #48 it seems.

Could you please do the following?
1. Rebase on top of `main`, resolve the conflicts.
2. Look for `// TODO[#48]` in the code and fix them.
3. Check if your code really solves #48.

Cesium.CodeGen/Extensions/TypeSystemEx.cs
```diff
@@ -15,27 +16,66 @@ internal static class TypeSystemEx
...
+
+    private static MethodInfo[]? FindMethodGroup(Assembly assembly, string typeName, string methodName)
+    {
+        return assembly.GetType(typeName)?.GetMethods().Where(x => x.Name == methodName).ToArray();
+    }
+
+    public static Type? GetTypeObject(this TypeReference typeReference)
+    {
+        return Type.GetType($"{typeReference.FullName},{typeReference.Scope}");
+    }
```
ForNeVeR, line 74: Unfortunately, this approach won't work: it tries to resolve types using reflection, and types aren't generally available in the compiler process (user types won't resolve via `Type::GetType`).

Why is it even required?

There's the same problem for most of your changes in this file: you're trying to manipulate with types and methods using runtime reflection, which isn't the right approach in a compiler.

ForNeVeR (changes requested):

Cesium.CodeGen/Extensions/TypeSystemEx.cs
ForNeVeR: The reflection code is already there, I'll need to think about it a bit more.

ForNeVeR (approved):
Thanks for help!

PR #157, opened 2022-07-30
Feature/usual arithmetic conversions

### LINKED ISSUE #136: Implement usual arithmetic conversions
We need to implement section "6.3.1.8 Usual arithmetic conversions" to support automatic conversions for binary operations (+, -, /, etc).

_- I think we need to finish #135 first and implement also corresponding rules for type propagation in expression. gsomix._

See `TODO[136]` in the code to find the clues for this.

impworks:
Closes #136.

Pointing out a few more things related to this logic, which should be done later but are out of scope for this particular task:

1. Arithmetic conversions currently work only in mathematical and bitwise binary operators. There are several more places in the language where values are also promoted, e.g. assignment and function invocation. This logic should be moved to `CastExpression` later to reuse it for explicit and implicit casts.
2. The parser does not support any numeric values except for `int` and `char`, so there are no tests. When it does, we should write tests that cover floats, doubles and longs.
3. Current implementation is inefficient for constants. For example, `1 + 'b'` gives the following:

```
IL_0000: ldc.i4 1
IL_0005: ldc.i4 98
IL_000a: conv.u1
IL_000b: conv.i4
IL_000c: add
IL_000d: stloc V_0
``` 

As you see, the char is first converted from int to byte, and then back to int. This could be removed if we implement lowering to implicit casts - the cast expression would see that the operand is a constant and emit it immediately. But this would require a lot of refactoring. so I left it for another day.


Cesium.CodeGen/Ir/Expressions/BinaryOperators/ComparisonBinaryOperatorExpression.cs
```diff
@@ -0,0 +1,61 @@
...
+            new ComparisonBinaryOperatorExpression(Left.Lower(), BinaryOperator.EqualTo, Right.Lower()),
+            BinaryOperator.EqualTo,
+            new ConstantExpression(new IntegerConstant("0"))
+        ),
+        _ => new ComparisonBinaryOperatorExpression(Left.Lower(), Operator, Right.Lower()),
+    };
+
+    public override void EmitTo(IDeclarationScope scope)
+    {
+        // todo: check if operand types are compatible?
```
ForNeVeR, line 45: Note for self: make an issue out of it.

ForNeVeR (approved):
Thanks!

PR #164, opened 2022-07-31
Optimized if-else when "true" branch always returns.

### LINKED ISSUE #113: `IfElseStatement`: optimize case of branches ending with `return`
See `// TODO[#113]` in the `Cesium.CodeGen.Ir.BlockItems.IfElseStatement`.

impworks:
Closes #113.


ForNeVeR (approved):
Thanks!

PR #171, opened 2022-08-02
More integration testing

kant2002:
Test that accessing nested fields work
I did not see that `.` access was tests in the structs somewhere
Also test second variant of forward declaraions


ForNeVeR (approved):
More tests = more fun. Thanks!

PR #180, opened 2022-08-06
Declare variables by IType, and not by VariableDef

kant2002:
That way we working more on C-types then Cecil internals

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
@@ -117,39 +117,39 @@ int main() { return foo(); }", "Function foo not defined.");
     public Task PrimitiveTypes() => DoTest(@"int main(void)
 {
     // basic
-    char c;
-    short s;
-    signed s1;
-    int i;
-    unsigned u;
-    long l;
-    float f;
-    double d;
+    char c = 0;
+    short s = 0;
+    signed s1 = 0;
+    int i = 0;
+    unsigned u = 0;
+    long l = 0;
+    float f = 0;
+    double d = 0;
 
     // unsigned
-    unsigned char uc;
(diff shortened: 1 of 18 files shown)


ForNeVeR (commented):
This is all nice, but we'll have to redesign several tests (the ones that were just defining local variables), since they have now stopped doing what they were supposed to.

kant2002:
What tests in your opinion require redesign?

kant2002:
Or more specifically, should I try preserve existing behavior?

ForNeVeR:
- `Cesium.CodeGen.Tests/verified/CodeGenMethodTests.FunctionPtrTest.verified.txt`
- `Cesium.CodeGen.Tests/verified/CodeGenMethodTests.FunctionPtrWithParamsTest.verified.txt`
- `Cesium.CodeGen.Tests/verified/CodeGenMethodTests.ImplicitReturnAllowedForMain.verified.txt`
- `Cesium.CodeGen.Tests/verified/CodeGenMethodTests.PrimitiveTypes.verified.txt`
- `Cesium.CodeGen.Tests/verified/CodeGenTypeTests.BasicTypeDef.verified.txt`
- `Cesium.CodeGen.Tests/verified/CodeGenTypeTests.TypeDefStructUsage.verified.txt`

I think it would be enough to move the declarations to global level in these tests to force-emit the fields for them.No need to preserve the old behavior.

ForNeVeR (approved):
Thanks!

kant2002:
What do you think if I add variable initializers to the variables. THat way corresponding variables would be used, and that's trigger local variables back.
Regarding moving type declarations on the global level. Because that trigger different path, I would consider this as separate set of test cases, which I gladly add if you think that's worth it.

ForNeVeR:
You did good.

PR #183, opened 2022-08-08
Add bunch of test cases

kant2002:
- Add both working, and not yet working snippets of C
- Add support for recursive dir traversal in integration testing. Will need this for more tests to C runtime

### DIFF
--- a/Cesium.Compiler/stdlib/stdio.h
+++ b/Cesium.Compiler/stdlib/stdio.h
@@ -1,2 +1,8 @@
 __cli_import("Cesium.Runtime.StdIoFunctions::PutS")
 void puts(char *s); // TODO[#156]: Change to int
+
+void printf(char* s) {
+    // That's temporary until varargs would not be implemented.
+    // no formatting obviously.
+    puts(s);
+}
--- a/Cesium.Compiler/stdlib/stdio.h
+++ b/Cesium.Compiler/stdlib/stdio.h
@@ -1,8 +1,5 @@
 __cli_import("Cesium.Runtime.StdIoFunctions::PutS")
 void puts(char *s); // TODO[#156]: Change to int
 
-void printf(char* s) {
-    // That's temporary until varargs would not be implemented.
-    // no formatting obviously.
-    puts(s);
-}
+__cli_import("Cesium.Runtime.StdIoFunctions::PutS")
+void printf(char* s);
--- a/Cesium.Runtime/StdLibFunctions.cs
+++ b/Cesium.Runtime/StdLibFunctions.cs
@@ -7,6 +7,7 @@ public unsafe static class StdLibFunctions
 {
     public static int Abs(int value)
     {
+        if (value == int.MinValue) return int.MinValue;
         return Math.Abs(value);
     }
 
--- a/Cesium.Compiler/Cesium.Compiler.csproj
+++ b/Cesium.Compiler/Cesium.Compiler.csproj
@@ -31,4 +31,8 @@
         <Content Include="stdlib/*" CopyToOutputDirectory="Always" />
     </ItemGroup>
 
+    <ItemGroup>
+      <None Remove="stdlib\math.h" />
+    </ItemGroup>
+
 </Project>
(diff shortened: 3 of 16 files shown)


kant2002:
At least 1 test should be removed after #185 lands, since I have basically duplicate test.

Cesium.Compiler/stdlib/stdio.h
```diff
@@ -1,2 +1,8 @@
 __cli_import("Cesium.Runtime.StdIoFunctions::PutS")
 void puts(char *s); // TODO[#156]: Change to int
+
+void printf(char* s) {
+    // That's temporary until varargs would not be implemented.
+    // no formatting obviously.
+    puts(s);
+}
```
ForNeVeR, line 8: I'm not sure if it is valid for `stdio.h` to define the function body in a header file.

Also, this deserves a TODO comment (I'm creating an issue for each merged `TODO`).

Cesium.Runtime/StdLibFunctions.cs
```diff
@@ -7,6 +7,7 @@ public unsafe static class StdLibFunctions
 {
     public static int Abs(int value)
     {
+        if (value == int.MinValue) return int.MinValue;
         return Math.Abs(value);
     }
```
ForNeVeR, line 12: Wow that's a gimmick. Let's write a unit test for the standard library, then?

ForNeVeR (changes requested):
Also, I'm not merging this until everything's green. Have you investigated the reason for the test failures? Do you need help?

Cesium.Runtime/StdLibFunctions.cs
kant2002: Maybe we can, but I mostly think that should do that as part of "acceptance testing", but in retrospection after I think a bit on this I do agree with you. That's subsystem should be tested.

Cesium.Compiler/stdlib/stdio.h
kant2002: Okay. that PR is kitchen sink of different failed attempts to find some place where I can have quick win. I will try to cleanup it which is quite possible, but still it reply on earlier RPs which I send, so I waiting for them to land. I will let you know when this would be ready for your attention.

kant2002:
Abs is parsing INT_MIN. Fixed.
printf is having float point constants as part of tree, If's in progress now

kant2002:
Ooh, abs is also requires nested `#include` because  I add `#include <math.h>`

Cesium.Compiler/stdlib/stdio.h
ForNeVeR: For such cases, you may consider marking a PR as a "draft". This way, you formally let me know that it is not yet ready to merge.

An example of a draft PR is #139.

kant2002: Yeah, I was lazy when I create it, but cannot convert PR to draft after creation. Maybe that's some permission issue, or require some setup?

ForNeVeR: Sorry, I'm not sure what the rules about drafts are, but I've found that I can convert it myself. So I did.

Let me know if you want to un-draft this, and cannot do it yourself.

kant2002:
Now it's waiting for #189 only

Cesium.Compiler/stdlib/stdio.h
ForNeVeR: What's your opinion on this `printf` implementation? Should we still merge it being in a header file?

kant2002: I would like to have `Cesium.Runtime` to be something minimal and provide as much implementation in C as possible. Example of functions which maybe written in C is `atoi`/`ioa`, we almost at the point where we can write them in C. Maybe that's require have something like `stdlib.c` which would be added automatically by compiler. That's just example, not something which I think though.

For marketing reasons, I think we should have `printf` even with limitation as serious as it is right now. Technically right now we don't have "Hello World".

ForNeVeR: My whole point is that in C, you cannot write code in headers.

If we want to write C stdlib (how do you distinguish it from `Cesium.Runtime`, by the way?), then we'll have to do it in C files, and link them somehow with the user's program.

For now, this is too much. Let's migrate this to the `Cesium.Runtime`, and create a separate issue to discuss details of implementation parts of the standard library in C.

ForNeVeR (approved):
Thanks!

PR #184, opened 2022-08-08
Parse INT_MIN value properly

kant2002:
Parse INT_MIN value properly
Right now I belive bug is in the tokenizer which parse `-34` as `-` (unary negation) and `34` (constant). That triggers bug in the lowering when we try to parse `-2147483648` and parse `2147483648` as integer, which is impossible.


ForNeVeR (approved):
Thanks!

PR #188, opened 2022-08-09
Basic define/ifdef/ifndef support

### LINKED ISSUE #70: Preprocessor macro support (`#define`, `#ifdef`)
We should support simple macro definitions and usage. Say, this should be valid:
```
#define foo main
#ifdef foo
  int foo() { return 0; }
#endif
```

Note that there's no fancy stuff (like `#if defined()`, or any non-trivial preprocessor expressions) in this issue, yet.

kant2002:
`#ifdef`/`#ifndef` control if subsequent tokens until `#endif` would be included.
Wiring passing defines from command line should be added separately

Closes #70


Cesium.Preprocessor/IDefinesContext.cs
```diff
@@ -0,0 +1,7 @@
+namespace Cesium.Preprocessor;
+
+public interface IDefinesContext
+{
+    bool TryGetDefine(string macro, out string? macroReplacement);
+    void Define(string macro, string? replacement);
+}
```
ForNeVeR, line 7: Let's call it `IMacroContext` and `TryResolveMacro` / `DefineMacro`?

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #189, opened 2022-08-12
Floatingpointconst

### LINKED ISSUE #187: Add support for floating constant
Cesium has floating data types, but not floating point constants.
Would be nice to add support for it.

Basically if start from 
https://github.com/ForNeVeR/Cesium/blob/ad0a6b386cdd78442c812437f6f37656ceb970c2/Cesium.CodeGen/Ir/Expressions/ConstantExpression.cs#L34
and create `FloatConstant` in similar wat to `IntegerConstant` we have it.

Fantoom:
closes #187

Added DoubleConstant for future use


Cesium.CodeGen.Tests/CodeGenTestBase.cs
```diff
@@ -10,6 +10,7 @@
 
 namespace Cesium.CodeGen.Tests;
 
+[UseCulture("")]
```
kant2002, line 13: I'm curious why explicit control of the culture is needed here? Isn't we should write code which runs on all cultures in same way.

Fantoom: [Instruction.ToString()](https://github.com/jbevain/cecil/blob/49b1c52df5e8d2bf03768dc35ba6575e88277bc7/Mono.Cecil.Cil/Instruction.cs#L95) uses the culture of the thread, which can lead to different results on different machines and fail tests

kant2002: Cool that you submit https://github.com/jbevain/cecil/pull/870 to fix underlying root cause. Would be interesting to know how it would be responded.

Cesium.CodeGen/Ir/Expressions/ConstantExpression.cs
```diff
@@ -34,6 +34,7 @@ private static IConstant GetConstant(Ast.ConstantExpression expression)
             CTokenType.IntLiteral => new IntegerConstant(constant.Text),
             CTokenType.CharLiteral => new CharConstant(constant.Text),
             CTokenType.StringLiteral => new StringConstant(constant),
+            CTokenType.FloatLiteral => new FloatConstant(constant.Text), //TODO: Add DoubleConstant
```
kant2002, line 37: As per C11 clause 6.4.4.2, “An unsuffixed floating constant has type double. If suffixed by the letter f or F, it has type float. If suffixed by the letter l or L, it has type long double.“

So I think we can distinguish between `float`, `double` and `long double` trivially.

Fantoom: In current implementation we don't have suffix support. I have added a todo comment to get back to it when it is supported

kant2002: Then probably would be good to have Double by default, at least that would be correct C behaviour, and if we miss `float` support, because of `f` suffix missing, then let's add `float` support later one.
Regarding missing suffix support, is this on us or on Yoakke?

Cesium.CodeGen.Tests/CodeGenMethodTests.cs
```diff
@@ -71,6 +71,7 @@ int main()
     [Fact] public Task Parameter1Get() => DoTest("int foo(int x){ return x + 1; }");
     [Fact] public Task Parameter5Get() => DoTest("int foo(int a, int b, int c, int d, int e){ return e + 1; }");
     [Fact] public Task CharConstTest() => DoTest("int main() { char x = '\\t'; return 42; }");
+    [Fact] public Task FloatConstTest() => DoTest("int main() { float x = 1.5; return 42; }");
```
kant2002, line 74: I think here would be good to have 
```
[Fact] public Task DoubleConstTest() => DoTest("int main() { double x = 1.5; return 42; }");
```
I think naming is from when Float was default. When we really implement float we need to change name anyway.

Fantoom: Yeah, I forgot about it

Cesium.CodeGen.Tests/verified/CodeGenMethodTests.FloatConstTest.verified.txt
```diff
@@ -0,0 +1,7 @@
+﻿System.Int32 <Module>::main()
+  Locals:
+    System.Single V_0
+  IL_0000: ldc.r8 1.5
+  IL_0009: stloc.0
+  IL_000a: ldc.i4.s 42
+  IL_000c: ret
```
ForNeVeR, line 7: There's no test corresponding to this gold file.

Cesium.CodeGen/Ir/Expressions/Constants/FloatConstant .cs
```diff
@@ -0,0 +1,27 @@
+using Cesium.CodeGen.Contexts;
```
ForNeVeR, line 1: The name of this file includes a space, please remove.

Cesium.CodeGen.Tests/UseCultureAttribute.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <param name="methodUnderTest">The method under test</param>
+    public override void After(MethodInfo methodUnderTest)
+    {
+        Thread.CurrentThread.CurrentCulture = originalCulture;
+        Thread.CurrentThread.CurrentUICulture = originalUICulture;
+
+        CultureInfo.CurrentCulture.ClearCachedData();
+        CultureInfo.CurrentUICulture.ClearCachedData();
+    }
+}
```
ForNeVeR, line 89: This seems to be copy-paste from xUnit repository: https://github.com/xunit/samples.xunit/blob/5dc1d35a63c3394a8678ac466b882576a70f56f6/UseCulture/UseCultureAttribute.cs

Copying external code is a little compilcated (we have to follow the license), and it also does a bit more work that I want it to (allows using named cultures etc.).

I suggest rewriting this attribute, making it only work with `InvariantCulture`, and removal of the copy-paste altogether (so there's only your code in the file).

Cesium.CodeGen/Ir/Expressions/ConstantExpression.cs
ForNeVeR: Suffix support is on us, because I want us to have a separate lexer eventually. Not sure if Yoakke already supports the suffices.

Cesium.CodeGen/Ir/Expressions/Constants/DoubleConstant.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Cesium.CodeGen.Ir.Expressions.Constants;
+
+internal class DoubleConstant : IConstant
+{
+    private readonly double _value;
+
+    public DoubleConstant(string value)
+    {
+        if (!double.TryParse(value, NumberStyles.Float, CultureInfo.InvariantCulture.NumberFormat, out _value))
+            throw new NotSupportedException($"Cannot parse an double literal: {value}.");
```
ForNeVeR, line 16: We use `CompilerException` here already :)

Cesium.CodeGen/Ir/Expressions/Constants/FloatConstant .cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Cesium.CodeGen.Ir.Expressions.Constants;
+
+internal class FloatConstant : IConstant
+{
+    private readonly float _value;
+
+    public FloatConstant(string value)
+    {
+        if (!float.TryParse(value, NumberStyles.Float, CultureInfo.InvariantCulture.NumberFormat, out _value))
+            throw new NotSupportedException($"Cannot parse an float literal: {value}.");
```
ForNeVeR, line 16: `CompilerException` here, too.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

ForNeVeR:
I've made an improvement of removing `FloatConstant` (it was of no use anyway) and replaced the `DoubleConstant` with a new `FloatingPointConstant`. My vision is that it is _likely_ that it'll serve for both `double` and `float`, depending on the expression type, though I'm not 100% sure on that.

The follow-up issue is #243.

PR #185, opened 2022-08-08
feat: implement getting address of function

FrediKats:
Try to close #181 

I am not sure about init getting LValueLocalVariable for func. FunctionInfo and FunctionType is not LValue. And... shouldn't be, is it?  But I don't know other way of implementation.

p.s. thx @kant2002 for FunctionType.

Issue #87, opened 2022-01-23
Functions with empty (non-void) parameter list


ForNeVeR:
Functions defined in this way has special meaning in C:
```
void foo() { }
```

This isn't the same as `void foo(void) {}` (which is a canonical empty parameter list).

Cesium should support those according to the standard.

PR #190, opened 2022-08-13
Call Exit runtime helper as part of exit procedure

kant2002:
That allow introduce `atexit` handling when function variables would be introduced.
Related #67


kant2002:
At exit need at least #185

ForNeVeR (changes requested):
Please note that you're only calling this function in the synthetic entrypoint, and not a normal entrypoint.

`int main(void)` is a "normal" entrypoint (i.e. it compiles as-is, and becomes an actual assembly entrypoint), while `int main(int, char*[])` cannot be declared an assembly entry point, and so a synthetic one (with a CLI-compatible signature, `int main(string[] args)`) gets created instead.

To properly handle `atexit`, this code should be emitted in both cases: for a normal and a synthetic entrypoint.

To handle them better, you may choose to emit it in an actual user-defined function `main()` (since it gets emitted equally in both cases), and not in a synthetic entrypoint code.

kant2002:
From my perspective compiler creates execution environment and inject user
code into it. Modifying main is modifying user code. But if we will create
syntetic entrypoint then user code is translated exactly, and we can create
execution environment as it fits to us

Вс, 14 авг. 2022 г. в 19:10, Friedrich von Never ***@***.***>:

ForNeVeR:
I think it's okay to always emit a synthetic entrypoint if we want `atexit` to always work (and we do).

kant2002:
So what my current understanding for syntetic entry point code
```
int exitCode;
try
{
   // Prepare parameters
   // Pin command line arguments if needed
   argC = args.Length;
   argV = Cesium.Runtime.RuntimeHelpers.ArgsToArgv(args);
   argVCopy = new byte*[argV.Length];
   fixed (byte** argVPtr = argVCopy)
   // Call user main
   exitCode = main(/* pass argC, argVPtr if needed*/);
}
catch (Exception ex)
{
   exitCode = ex.HResult;
}

Environment.Exit(exitCode);
return exitCode;
```
In addition to synthetic entry point we will need subscribe to app domain exit in module initialzer with subscription to `AppDomain.ProcessExit`. will add that later, when we have actual `atexit` support

ForNeVeR:
I don't think we have to to `try … catch` there. Just let the runtime to fail by itself. Maybe that'll give better error handling than anything we can write in C currently.

kant2002:
I'm not sure about that one. I do not know, but that's passive position probably would bite us in the future. I honestly do not have strong arguments for `try/catch`, but I did not put a lot of time in design as well.
What I have in my mind currently.
Let's say end-user (developer) wrote program and she want to control how errors are reported by application. What if we have buggy interop function which sometimes crashes. That immediately escalate and lead to program crash. C do not have ability put giant `try/catch` around program, so we are lacking of options. 

That's honestly interop related, and this is even more complicated, but probably we should continue talking about options.

Even simply adding synthetic entrypoint for other forms of `main` add much of a noise (you will see it yourself), and I would like reduce churn in tests by designing things where possible.

Cesium.CodeGen.Tests/CodeGenMethodTests.cs
```diff
@@ -61,9 +61,9 @@ int main()
 
     [Fact] public Task ParameterlessMain() => DoTest("int main(){}");
     [Fact] public Task VoidParameterMain() => DoTest("int main(void){}");
+    [Fact] public Task VoidReturnMain() => DoTest("void main(void){}");
```
ForNeVeR, line 64: The standard doesn't support this, and I see no reason for us to do it.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
```diff
@@ -44,10 +44,6 @@ public void EmitTo(TranslationUnitContext context)
     {
         var (parameters, returnType) = _functionType;
         var resolvedReturnType = returnType.Resolve(context);
-        if (IsMain && resolvedReturnType != context.TypeSystem.Int32)
-            throw new CompilationException(
-                $"Invalid return type for the {_name} function: " +
-                $"int expected, got {returnType}.");
```
ForNeVeR, line 50: I suggest to return this check back.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
```diff
@@ -122,22 +129,15 @@ private bool ValidateMainParameters()
...
-    /// <returns>A synthetic entrypoint method created.</returns>
-    private MethodDefinition GenerateSyntheticEntryPoint(
+    private MethodDefinition GenerateSyntheticEntryPointStrArray(
         TranslationUnitContext context,
         MethodReference userEntrypoint)
     {
-        var module = context.Module;
         var syntheticEntrypoint = new MethodDefinition(
-            "<SyntheticEntrypoint>",
+            "<SyntheticEntrypoint>StrArray",
```
ForNeVeR, line 140: _(Minor)_ Do we have to use different function names for different `main` variants? I don't think so, but I'm ready to listen to the arguments.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
```diff
@@ -220,6 +221,47 @@ private bool ValidateMainParameters()
...
+        MethodReference userEntrypoint)
+    {
+        var syntheticEntrypoint = new MethodDefinition(
+            "<SyntheticEntrypoint>",
+            MethodAttributes.Public | MethodAttributes.Static,
+            context.TypeSystem.Int32)
+        {
+            Parameters =
+            {
+                new ParameterDefinition("args", ParameterAttributes.None, context.TypeSystem.String.MakeArrayType())
```
ForNeVeR, line 241: For the "simple" variant, we don't need the arguments, do we?

Cesium.Parser.Tests/PreprocessorTests/verified/PreprocessorTests.IncludeDefine.verified.txt
```diff
@@ -0,0 +1,4 @@
+﻿#define XXX 123
+int test()
+{
+}
```
ForNeVeR, line 4: Are you sure we need this file? Looks like a leftover from another branch.

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
kant2002: Technically I suspect that MSVC team do this for some reasons, but let's keep all syntetic entry points as close as possible. There even slightest idea right now in my mind.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
kant2002: not sure, #87 right now handle with "simple" signature, but I'm looking at unification of "simple" and array in some way. At least this is my current understanding.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
ForNeVeR: Well, that would be an argument, but whatever: you've already changed the name back. Okay!

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
ForNeVeR: Let's either do this unification, or drop the `args` from here :)

Also, even if you unify the code, I don't think that `args` will contribute to that.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
kant2002: We don’t know and better not have hint to ancient knowledge like this. Most likely we forget about it. And if that’s really needed we definitely hit some issue and can make change

ForNeVeR: Agree. Let the ancient ones keep their secrets.

Cesium.CodeGen/Ir/TopLevel/FunctionDefinition.cs
kant2002: I'm not in mood for unification. Would like to sleep well, knowing that this PR lands. So I remove `args` from `Simple`

ForNeVeR: No problemo.

ForNeVeR (approved):
Thanks!

PR #193, opened 2022-08-13
Add support for negation in expressions

kant2002:
I would like to simplify that by changing model for IfElseStatment to have differnet mode for branching, but that's require changes in lowering, and I feel that's too complicated

Related to #197


ForNeVeR (approved):
Thanks!

PR #194, opened 2022-08-14
Allow place integration tests in subdirectories

kant2002:
That allow have more tests in more structured manner


ForNeVeR (approved):
Thanks!

PR #209, opened 2022-08-14
Exceptions and documentation

### LINKED ISSUE #204: Migrate exceptions to the modern model
See #199.

All the current exceptions (if appropriate) should migrate to the new model.

`WipException` is a bit tricky, so I suggest adding a `WipException.Todo` constant with a value of `-1` which we'll use in ase when the issue number is unknown.

ForNeVeR:
Closes #204.

**TODO:**
- [x] refactor the parser-related issues
- [x] finish the moving to the new exception system
- [x] status update on the compiler and parser support
- [x] review all the `TODO` in the project and connect to the corresponding issues as well
- [x] review the created issues and make sure no `xxx` is remaining there
- [x] address the PR feedback
- [x] review


Cesium.Core/Exceptions/WipException.cs
```diff
@@ -0,0 +1,17 @@
+namespace Cesium.Core.Exceptions;
```
kant2002, line 1: Can we have exceptions in Cesium namespace. Otherwise eventually we will have `using Cesium.Core.Exceptions;` everywhere. And I mostly do not see a reason why we would not need them.

Cesium.Core/Cesium.Core.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
kant2002, line 1: If we have Core, then `PreprocessorException` should be part of this exception infrastructure. We definitely would have some tricky preprocessor situations.

```
#define EOF 1
#define VALIDATE_EOF(code) if (code == EOF) { abort(-1); }
```

Most likely this will force us to store preprocessor AST as macro definitions and expand them. `VALIDATE_EOF` has `code` parameter which can be some other macro itself, and we probably should build tree and materialize it. I most likely would like to have WIP exception here.

Cesium.Core/Exceptions/WipException.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    public WipException(int issueNo, string additionalMessage)
+        : base($"This work is in progress. {additionalMessage}. See https://github.com/ForNeVeR/Cesium/issues/{issueNo}.")
+    {
+    }
+
+    public WipException(int issueNo)
+        : base($"This work is in progress. See https://github.com/ForNeVeR/Cesium/issues/{issueNo}.")
+    {
+    }
```
kant2002, line 16:
```suggestion
    }
    
    public static WipException Unknown(string message) => new WipException(ToDo, message);
```

docs/exceptions.md
```diff
@@ -0,0 +1,15 @@
...
+
+All the Cesium-produced exceptions are derived from `Cesium.Core.Exceptions.CesiumException`. Any other exceptions thrown from the compiler code should be treated as critical internal compiler exceptions.
+
+- `Cesium.Core.Exceptions.ParseException` should be thrown if we were unable to parse something.
+- `Cesium.Core.Exceptions.CompilationException` should be thrown if an error happened during compilation (i.e. a type or a variable was not found): this is most likely due to the user code being incorrect.
+- `Cesium.Core.Exceptions.AssertException` is thrown when the compiler code failed its own assertions. Most likely, it means a bug in the compiler.
+- `Cesium.Core.Exceptions.WipException` is an exception which gets thrown from areas where certain features aren't implemented, but they will sometime (ideally, at least). Every such exception should be linked with an issue on GitHub.
+
+  When writing new code, please use `throw new WipException(WipException.Unknown, "some optional description")`. The project maintainers will convert these to the issues while merging your contribution.
```
kant2002, line 13:
```suggestion
  When writing new code, please use `throw new WipException(WipException.ToDo, "some optional description")`. The project maintainers will convert these to the issues while merging your contribution.
```

kant2002: Also look for sibling comment with proposal to have helper method for this kind of situations.

Cesium.Core/Exceptions/WipException.cs
ForNeVeR: Agree on this point.

Cesium.Core/Exceptions/WipException.cs
ForNeVeR: Nope. I prefer people to explicitly mention `WipException.ToDo` in their code, for easier monitoring on the review stage.

Cesium.Core/Cesium.Core.csproj
ForNeVeR: Addressed.

ForNeVeR:
@kant2002, thanks for your review!

PR #247, opened 2022-08-19
Support basic #define replacement

kant2002:
I do not account for `#define TEST(X) X` cases for example.

Contributes to #70


Cesium.Preprocessor/CPreprocessorTokenType.cs
```diff
@@ -21,6 +21,15 @@ public enum CPreprocessorTokenType
...
     PreprocessingToken,
+
+    [Regex("[;+-*/]+")]
+    Separator,
+
+    [Token("(")]
+    LeftParent,
+
+    [Token(")")]
+    RightParent,
```
ForNeVeR, line 34:
```suggestion
    LeftParen,

    [Token(")")]
    RightParen,
```

ForNeVeR (changes requested):

Cesium.Preprocessor/CPreprocessorTokenType.cs
kant2002: You catch me without my pants :)

ForNeVeR (approved):
Thanks!

PR #249, opened 2022-08-20
Allow have multiple declarations of same type

kant2002:


Cesium.CodeGen.Tests/CodeGenMethodTests.cs
```diff
@@ -108,6 +108,20 @@ public void IncorrectOverrideCliImport() => DoesNotCompile(@"__cli_import(""Syst
...
+int console_read(void);");
+
+    [Fact]
+    public void CanHaveTwoFucntionDeclarations() => DoTest(@"
+int console_read(void);
+
+int console_read(void);
+
+int console_read(void) { return 0; }");
+
```
ForNeVeR, line 124: Let's also add a test that doesn't compile (if the function signatures of `__cli_import` clauses are different)?

Cesium.Compiler/stdlib/limits.h
```diff
@@ -0,0 +1,34 @@
...
+#define _I16_MAX      32767i16
+#define _UI16_MAX     0xffffui16
+
+#define _I32_MIN    (-2147483647i32 - 1)
+#define _I32_MAX      2147483647i32
+#define _UI32_MAX     0xffffffffui32
+
+#define _I64_MIN    (-9223372036854775807i64 - 1)
+#define _I64_MAX      9223372036854775807i64
+#define _UI64_MAX     0xffffffffffffffffui64
```
ForNeVeR, line 34: _(Note)_ This shouldn't be a part of this review, but I'm fine with merging this improvement here.

Cesium.Compiler/stdlib/limits.h
```diff
@@ -0,0 +1,34 @@
...
+#define USHRT_MAX     0xffff
+#define INT_MIN     (-2147483647 - 1)
+#define INT_MAX       2147483647
+#define UINT_MAX      0xffffffff
+#define LONG_MIN    (-2147483647L - 1)
+#define LONG_MAX      2147483647L
+#define ULONG_MAX     0xffffffffUL
+#define LLONG_MAX     9223372036854775807i64
+#define LLONG_MIN   (-9223372036854775807i64 - 1)
+#define ULLONG_MAX    0xffffffffffffffffui64
```
ForNeVeR, line 18: Sorry, what's `i64`? Is it a valid C17 syntax?

ForNeVeR (changes requested):

Cesium.Compiler/stdlib/limits.h
kant2002: It seems MSVC is too permissive in what it allows. Will read docs and change accordingly.

Cesium.Compiler/stdlib/limits.h
kant2002: I add limits.h because Ubuntu build was failing without that. So I did not see reason to include just 2 definitions here

ForNeVeR (approved):
Thanks!

PR #251, opened 2022-08-20
Add AOT support

kant2002:
- Provide parameters to specify paths to S.P.CoreLib and Cesium.Runtime
- Provide rd.xml with only class which require keep everymetadata.

If target net7.0 project can be published using NativeAOT.
During running compiler only additional switch which should be used is `--corelib` because we cannot rely on reflection to find existing CoreLib assembly in the process.


Cesium.Compiler/Arguments.cs
```diff
@@ -1,39 +1,48 @@
...
+    [Option("import", HelpText = "Provides path to assemblies which would be added as references automatically into resulting executable.")]
+    public IList<string> DefaultImportAssemblies { get; init; } = null!;
+
+    [Option("corelib", HelpText = "Sets path to CoreLib assembly")]
+    public string? CoreLib { get; init; }
+
+    [Option("runtime", HelpText = "Sets path to Cesium C Runtime assembly")]
+    public string? CesiumCRuntime { get; init; }
+
+}
```
ForNeVeR, line 48: Sorry, but you seem to have problems with `\r\n` again :(

I recommend you to set `git config core.autocrlf input`

Cesium.Compiler/Compilation.cs
```diff
@@ -17,12 +17,15 @@ internal static class Compilation
...
         TargetRuntimeDescriptor targetRuntime,
-        ModuleKind? moduleKind = null,
-        string @namespace = "",
-        string globalClassFqn = "")
+        ModuleKind? moduleKind,
+        string corelibAssembly,
+        string cesiumRuntime,
+        IList<string> defaultImportAssemblies,
+        string @namespace,
+        string globalClassFqn)
```
ForNeVeR, line 25: Okay, it quickly becomes a lot of parameters. Let's move them into a `record`? What do you think?

Cesium.Compiler/Main.cs
```diff
@@ -1,45 +1,48 @@
...
+
+static void DisplayHelp<T>(ParserResult<T> result)
+{
+    var helpText = HelpText.AutoBuild(result, h =>
+    {
+        h.AddEnumValuesToHelpText = true;
+        return HelpText.DefaultParsingErrorsHandler(result, h);
+    }, e => e);
+    Console.WriteLine(helpText);
+}
```
ForNeVeR, line 48: The problem with `\r\n` here, too.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #252, opened 2022-08-20
Add support of complex subscript expressions

### LINKED ISSUE #230: Subscript expressions (aka index operator) of complex expressions
As of now, we don't seem to support something like this:
```c
typedef struct { int[10] x; } foo;
int main() {
  foo a;
  a.x[0] = 0;
}

Look for the number `230` in the code to find clues to implement this feature.

### PR DESCRIPTION
**Pull request by te9c**
Closes #230.

### CODE REVIEW

File: Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs [hunk 1]
```diff
@@ -35,7 +35,12 @@ public IExpression Lower()
     public IValue Resolve(IDeclarationScope scope)
     {
         if (_expression is not IdentifierExpression identifier)
-            throw new WipException(230, "Subscription supported only for IdentifierConstantExpression");
+        {
+            if (_expression is not IValueExpression IValueExpresssion)
```
kant2002, line 39:
```suggestion
            if (_expression is not IValueExpression valueExpresssion)

```

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
```diff
@@ -35,7 +35,12 @@ public IExpression Lower()
     public IValue Resolve(IDeclarationScope scope)
     {
         if (_expression is not IdentifierExpression identifier)
-            throw new WipException(230, "Subscription supported only for IdentifierConstantExpression");
+        {
+            if (_expression is not IValueExpression IValueExpresssion)
+                throw new CompilationException($"{_expression} is not IValueExpression");
```
kant2002, line 40: This message would be visible to user (developer who wrote program). It is impolite expect that developer know our compiler in such details. Probably would be better if we have message like this
```suggestion
                throw new CompilationException($"{_expression} is not a value expression");

```
Also would be good to have tests on this functionality.

kant2002: My message is probably off, but for other reasons. I cannot imagine better expression just now. Maybe @ForNeVeR come up with better idea, or after we have tests something pop into my mind.

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
```diff
@@ -35,7 +35,12 @@ public IExpression Lower()
     public IValue Resolve(IDeclarationScope scope)
     {
         if (_expression is not IdentifierExpression identifier)
```
ForNeVeR, line 37: `IdentifierExpression` is also an `IValueExpression`, too, so you may remove the first type check.

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
```diff
@@ -35,7 +35,12 @@ public IExpression Lower()
...
-            throw new WipException(230, "Subscription supported only for IdentifierConstantExpression");
+        {
+            if (_expression is not IValueExpression IValueExpresssion)
+                throw new CompilationException($"{_expression} is not IValueExpression");
+
+            return IValueExpresssion.Resolve(scope);
```
ForNeVeR, line 42: Please add a new test verifying this functionality. Take a look at [the testing guide](https://github.com/ForNeVeR/Cesium/blob/0b8a10d10c898c190463a43eb3053913d9f4ca63/docs/tests.md) and let me know if it's unclear how to proceed.

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
te9c: But if we dont have the first check, method will always return `IValueExpresssion.Resolve(scope);` and not the `LValueArrayElement`.

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
te9c: Ok! For now everything looks clear.

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
ForNeVeR: Ah, you're right. Let's clarify the logic a bit then?
Something like
```csharp
switch (_expression)
{
  case IdentifierExpression identifier: …
  case IValueExpression valueExpression: …
  default: throw new CompilationException(…);
}
```

te9c: Yep. Thats looks better

Cesium.IntegrationTests/complex_subscript_expressions.c
```diff
@@ -0,0 +1,7 @@
+typedef struct { int x[10]; } foo;
```
kant2002, line 1: It seems to be we have test for this https://github.com/ForNeVeR/Cesium/blob/main/Cesium.IntegrationTests/struct_subscription.ignore.c
If you can remove `.ignore` it will start running.

kant2002: Also I assume @ForNeVeR ask for new test somewhere in the [Cesium.CodeGen.Tests](https://github.com/ForNeVeR/Cesium/tree/main/Cesium.CodeGen.Tests) project

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
```diff
@@ -31,12 +31,14 @@ public IExpression Lower()
     public void EmitTo(IDeclarationScope scope) => Resolve(scope).EmitGetValue(scope);
 
     public IType GetExpressionType(IDeclarationScope scope) => Resolve(scope).GetValueType();
-
+
```
kant2002, line 34: nit: can you remove this?

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
```diff
@@ -31,12 +31,14 @@ public IExpression Lower()
...
     public IValue Resolve(IDeclarationScope scope)
     {
-        if (_expression is not IdentifierExpression identifier)
-            throw new WipException(230, "Subscription supported only for IdentifierConstantExpression");
-
-        return new LValueArrayElement(identifier.Resolve(scope), _index);
+        return _expression switch
+        {
+            IdentifierExpression identifier => new LValueArrayElement(identifier.Resolve(scope), _index),
+            IValueExpression valueExpression => valueExpression.Resolve(scope),
```
ForNeVeR, line 40: Now wait a minute, something's missing here. It looks like the the `_index` is ignored in this case? It shouldn't be.

And please add a unit test (to, say, `CodeGenMethodTests` class).

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
ForNeVeR: To be honest, I don't care about the formatting _that_ much.

Cesium.CodeGen.Tests/verified/CodeGenMethodTests.StructSubscription.verified.txt
```diff
@@ -0,0 +1,21 @@
...
+  Locals:
+    foo V_0
+  IL_0000: ldloca.s V_0
+  IL_0002: conv.u
+  IL_0003: ldc.i4.0
+  IL_0004: stfld foo/<SyntheticBuffer>fixedArr foo::fixedArr
+  IL_0009: ldloca.s V_0
+  IL_000b: conv.u
+  IL_000c: ldfld foo/<SyntheticBuffer>fixedArr foo::fixedArr
+  IL_0011: ret
```
ForNeVeR, line 11: Now you can see that this doesn't include `[3]` anywhere. This is a problem.

ForNeVeR (changes requested):

Cesium.CodeGen.Tests/verified/CodeGenMethodTests.StructSubscription.verified.txt
kant2002: @te9c do you need some help with finishing this?

te9c: @kant2002 yes, i would be grateful, just don't understand how we can return array element from field?

kant2002: @te9c I think after https://github.com/ForNeVeR/Cesium/pull/273 lands, you can continue poking at this PR after you rebase on main.

kant2002: @te9c can you rebase/merge on main and try move forward?

kant2002:
@te9c it's Hacktoberfest in the air. Are you willing to finish this PR in the near future?

Cesium.CodeGen.Tests/CodeGenMethodTests.cs
```diff
@@ -296,4 +296,18 @@ public void ImplicitReturnDisallowedNonMain() => DoesNotCompile(@"int foo()
 {
     int unused;
 }", "Function foo has no return statement.");
+
+    [Fact]
+    public Task StructSubscriptionTest() => DoTest(@"typedef struct
+{
+    int fixedArr[4];
```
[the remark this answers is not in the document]
ForNeVeR: In C, this struct has not a pointer field, but a special place for 4 `int` objects embedded directly into the structure body. To specify a type for such an unusual field, we generate a special type called `<syntheticBuffer>` (an unusual name to prevent the user from accessing it in their code directly). This is more or less in line with what Roslyn does for a `fixed` array.

ForNeVeR:
@te9c, thanks for your contribution! The changes have been finally merged in scope of my bigger PR #499.

PR #253, opened 2022-08-21
Document publish process for Cesium

kant2002:
This produces executable which can be packed into SDK + tools Nugets.

I move version management to shared `Directory.Build.props` so versioning would apply to all projects.
Also I use `VersionPrefix` so it would be easier for us to add custom suffix in automated fashion. For example add `--version-suffix ci-123` will produce `0.0.1-ci-123` version.

Bump version to 0.0.1, for SDK+tools I would like to have at least some version.


ForNeVeR (approved):
Thanks!

Issue #80, opened 2022-01-16
.NET SDK


ForNeVeR:
I want the users to be able to develop Cesium projects using .NET SDK, with a workflow like this:
1. Create a `.cesproj` file similar to this:
   ```xml
   <Project Sdk="Cesium.Sdk/1.0.0">
     <PropertyGroup>
       <TargetFramework>net6.0</TargetFramework>
     </PropertyGroup>
   </Project>
   ```
2. Run `dotnet build`: it should download the `Cesium.Sdk` from NuGet, set up compiler target, probably even download the standard library from NuGet!

In other words, Cesium projects should be fully compatible with normal .NET ecosystem; ideally, it should even be possible to use `<ProjectRefrence Include="../something.cesproj" />` in a C# project, and get the reference!

Depends on:
- [x] #161

kant2002:
This is how I done F* SDK https://github.com/kant2002/fstarmsbuildsdk . Tools packaging is based on how Zig handle that. I just borrow simplified version of it. Cesium has ability package itself as MSBuild task, so some variations possible.

ForNeVeR:
Wow, yeah, your F* SDK would really help here. I didn't think about it myself.

PR #254, opened 2022-08-21
Publishing of the compiler using MSBuild

kant2002:
This is same as #253 (and require changes in proj file from it) but for publishing
we using more modest `dotnet build . /t:PublishApp -c Release`
If expanding targets, we can ZIP, create SDK Nugets and perform other tasks

### DIFF
--- a/Cesium.IntegrationTests/Cesium.IntegrationTests.proj
+++ b/Cesium.IntegrationTests/Cesium.IntegrationTests.proj
@@ -5,6 +5,17 @@
         <Content Include="Run-Tests.ps1" />
     </ItemGroup>
 
+    <!-- This allow run restore on project when use MSBuild task without warning -->
+    <Target Name="_IsProjectRestoreSupported"
+          Returns="@(_ValidProjectsForRestore)">
+
+        <ItemGroup>
+            <_ValidProjectsForRestore Include="$(MSBuildProjectFullPath)" />
+        </ItemGroup>
+    </Target>
+    <Target Name="Restore" />
     <Target Name="Build" />
     <Target Name="VSTest" />
+
+    <Import Project="../Directory.Build.targets" />
 </Project>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,0 +1,16 @@
+<Project>
+    <Target Name="PublishApp" Condition="'$(MSBuildThisFileDirectory)Cesium.Compiler\Cesium.Compiler.csproj' == '$(MSBuildProjectFullPath)'">
+        <PropertyGroup>
+            <_ResultDir>$(MSBuildThisFileDirectory)artifacts\bin\Compiler\</_ResultDir>
+            <TargetRuntimeIdentifier>win-x64</TargetRuntimeIdentifier>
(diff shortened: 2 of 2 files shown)


ForNeVeR:
To me, it's unclear why do we need this. Could you please elaborate?

kant2002:
I expect that we have approximate following additional work in order to support #80. 
I see SDK as split between SDK + set of runtime tools packages. runtime tools is actual self-contained cesium shipped as Nuget package. Each platform will require their own executable.
Basically how it was done by Zig and  FStar is unpack ZIP into `tools` folder of runtime package. From that point dotnet package nuget for you.
So basically we will have 3 distribution packages
- Cesium compiler
- Cesium SDK
- Cesium-as-nuget

Cesium-as-nuget depends on Cesium compiler artifact, and somebody have to orchestrate shipping Cesium compiler artifact to proper place. Simple dotnet build would not be enough.

Once 3 pieces is done, we can use Cesium via simple `dotnet run`

ForNeVeR:
My vision is a bit different. We should have three parts.
1. Cesium compiler (in a NuGet package), maybe in the form of a library and not even directly executable by the end user.
2. Cesium runtime library (a normal NuGet), in future there may be multiple flavors of it.
3. Cesium SDK, which brings and uses a compiler NuGet (and maybe optionally a runtime library, similar to how FSharp.Core is delivered by F# SDK).

In your schema, I'm not getting what's the difference between Cesium compiler and Cesium-as-nuget. Also, the naming of "the runtime tools" confuses me. Is it the same as "Cesium compiler"? I can't see why we would want to repack it for each platform.

And, in any case, disregarding the discrepancies between our visions, I can't see how this `/t:PublishApp` is better than the ol' good `dotnet publish`. Could you please explain the possible usage of that?

kant2002:
I think we have only minor differences in vision. All of them stems from the fact that I by default think that compiler bits would be self-contained, but you that bits would be framework-dependent. We definitely should discuss that.

`/t:PublishApp` is better only if we need slightly more complicated build. If we have framework dependent distribution, that most likely would not be needed. If we will have Cesium as self-contained executable then we probably should move bits here and there, and that's require something separate. Maybe we can have separate publishing project. For me that's no different from custom target.

ForNeVeR:
To note: I'm not against publishing Cesium as a self-contained or an AOT-compiled executable in some form, this just doesn't relate to my vision of Cesium SDK (or at least not one of the immediate goals that should be solved while developing the SDK).

Regarding this `/t:PublishApp`, I am still not getting the proposed usage. At what point should it be called?

I understood the `dotnet publish` part since you have that documented. But this `/t:PublishApp` will just sit there in the code for now, for no known purpose. Just some code we'll have to support and the maintainer doesn't understand it.

kant2002:
Let this PR stay here for a bit. You may convert it to draft, I will think about options how we can gradually move to place where one way of the other would be more clearly visible.

One way to move in that direction is to create distribution when we hit first milestone, so people will use cesium standalone, and will see from that point on.

ForNeVeR:
Let's instead open a separate issue and discuss what artifacts and in which form we would like to publish from the compiler, and maybe set up the corresponding CI rules, if you wanna do something with it anyway?

Issue #127, opened 2022-06-22
Int128 support


ForNeVeR:
I think that it would be fun to have `long long` of size 128 bit. .NET 7 includes `System.Int128`. Let's consider making `long long` as `Int128` on .NET 7?

Perhaps this will require us to either switch to .NET 7 or provide a separate variant of stdlib for that, this is okay I guess.

GitMensch:
Please don't do that - there is int128 (even if it is an extension) in C, this is what you should map to Int128.

ForNeVeR:
Please tell me why not? What's the mind set of the person who uses `long long` in their C code but don't want it to be bigger than just `long`?

GitMensch:
C programmers want to have it longer than an `int` :-) [this is all related to decades of history)
For longer types they use [`__int128`](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html).

And of course, this would break the ABI into native modules :-)

ForNeVeR:
Since this is a contradictory idea (I have to admit), we will likely add an option to disable it.

When talking about ABI boundaries, I believe more options would be needed, such as some kind of "ABI profiles" (to make `long` shorter when interoprating with MSVC-compiled code). We'll see about that.

PR #255, opened 2022-08-21
Add support for #else directive

### LINKED ISSUE #250: Add support for #ifdef/#else/#endif
Right now `#ifdef/#endif` supported, but you cannot include `#else`.

kant2002:
Extend limits with signed/unsigned char options. By default MSVC/GCC/LLVM use signed char and you can enable unisnged char if you like using switch.

We can expose `-funsigned-char` switch using command line option afterwards.

Closes #250

### DIFF
--- a/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
+++ b/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
@@ -77,6 +77,27 @@ int foo() { return 0; }
 ");
 
     [Fact]
+    public Task IfElseLiteral() => DoTest(
+@"#define foo main
+#ifndef foo
+int myfoo() { return 0; }
+#else
+int foo() { return 0; }
+#endif
+");
+
+    [Fact]
+    public Task NestedIfDefined() => DoTest(
+@"#define foo main
+#ifdef foo
+int foo() { return 0; }
+#ifdef xfoo
+int foo() { return 0; }
+#endif
+#endif
+");
+
+    [Fact]
     public Task IfNotDefinedLiteral() => DoTest(
 @"#define foo main
 #ifndef foo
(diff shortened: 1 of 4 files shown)


ForNeVeR:
To be honest, I can't see why we need such an option. Why not just stick with the standard (which allows us to do an arbitrary choice and keep it)?

Cesium.Compiler/stdlib/limits.h
```diff
@@ -3,6 +3,14 @@
 #define SCHAR_MAX     127
 #define UCHAR_MAX     0xff
 
+#ifndef CESIUM_CHAR_UNSIGNED
```
ForNeVeR, line 6: To be honest, I can't see why we need such an option. Why not just stick with the standard (which allows us to do an arbitrary choice and keep it)?

Supporting this will be a headache. I'd prefer us to stick with the unsigned variant for now, and ditch the signed one.

Too many compiler options (which we have to support, no less!) would be bad at this point of Cesium lifetime.

ForNeVeR (changes requested):

Cesium.Compiler/stdlib/limits.h
kant2002: By default char is signed on major platforms. I suspect that would be headache to compile code with deviation from what most compilers expect.

I know that this is speculative change, because I add it just to test `#if/#else/#endif`. So can remove.

ForNeVeR: One of the minor purposes of Cesium is to invent a "better" (in some terms, ha-ha) C implementation: say, we have `long` and `int` different which is not in line with some other implementations.

I think that making `char` unsigned (and thus more in line with `System.Byte`) is one of these points. Interestingly, we even have this documented: `char` maps to `System.Byte` and not `System.SByte`.

In the future, we may try changing the decision if it shows itself problematic, but for now I'd keep it as-is. `char` should be unsigned for now.

ForNeVeR: One other absolutely heretical thought I'd like us to go along with is [having `long long` an actual 128-bit type](https://github.com/ForNeVeR/Cesium/issues/127) instead of what everyone else is doing :)

ForNeVeR (approved):
Thanks!

Issue #173, opened 2022-08-02
Assigning to dereferenced pointer breaks


kant2002:
```
int main(int argc, char *argv[])
{
    int a[10];
    int* b = &a[1];
    *b = 42;
    return a[1];
}
```

PR #258, opened 2022-08-24
Unignore working test

kant2002:
Quite likely was fixed in https://github.com/ForNeVeR/Cesium/pull/257

Fixes: #173


ForNeVeR (approved):
Thanks!

Issue #132, opened 2022-07-16
Architecture-dependent sizes in portable code


ForNeVeR:
There's a problem: if we compile to `AnyCPU`, then we cannot generally determine `sizeof(void*)`, which has to be a constant expression, according to the standard. Same about, say, `size_t` or `ptrdiff_t`, and actually any architecture-dependent types, mostly derived from pointers.

There are two possible strategies of dealing with that.

1. We can forbid using `sizeof` with a pointer and any pointer-sized types in portable code. Our users will have to specify the architecture when compiling a binary that uses these features. 

   It is possible that this will effectively prevent compiling to `AnyCPU` of the most of C code.

   Also, using of such C libraries from external code would be problematic sometimes (the library author will have to distribute versions built for each architecture, which kinda loses the point of Cesium).
2. We can introduce a special architecture-independent compilation mode which will use 8 bytes for any pointer or `size_t`, and generally will prefer bigger object sizes, but will still allow building to `AnyCPU`.

   This may be problematic for cases when a C library exposes some .NET interface which operates on pointer types of `IntPtr`: it's unclear how to properly compile that (or if it is even possible).

For now, I am considering going _both_ ways: add a flag to enable (or disable) architecture-independent compilation, and allow the user to specify the architecture that will be used to calculate `sizeof` and whatnot.

kant2002:
Basically what I propose is instead of `int IType.SizeInBytes` have `IExpression IType.GetSizeInBytes()` function or property. If paired with constant folding, we have same IL as today, and we can have dynamic IL for types like `void*[10]`

ForNeVeR:
Ok, after some thought, I've decided we'll try to support the following architecture sets in Cesium:
- `32b`: an architecture with 32-bit pointers, gets compiled as an x86 assembly (maybe with a flag to make it compatible with ARM32, though I'm not sure such a flag exists) 
- `64b`: an architecture with 64-bit pointers, gets compiled as an x64 assembly (with obligatory ARM64 support)
- `wide`: an architecture forcing pointers to be 64-bit even on a 32-bit platform (we'll have a lot of fun with that I reckon)
- `dynamic`: an architecture calculating pointer size dynamically at runtime

**Notes:**
[...]

PR #261, opened 2022-08-24
Implement pointer addition

kant2002:


Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
```diff
@@ -31,9 +30,21 @@ public override void EmitTo(IDeclarationScope scope)
 
         Left.EmitTo(scope);
         EmitConversion(scope, leftType, type);
+        if (rightType is PointerType rightPointerType)
+        {
+            scope.Method.Body.Instructions.Add(Instruction.Create(OpCodes.Ldc_I4, scope.CTypeSystem.GetTypeSize(rightPointerType.Base)));
+            EmitConversion(scope, scope.CTypeSystem.Int, type);
```
ForNeVeR, line 36: Now wait a minute. Are you converting a pointer to an `Int32` here? This is problematic!

I suggest introducing a special conversion to `nint` here.

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
```diff
@@ -49,6 +60,33 @@ public override IType GetExpressionType(IDeclarationScope scope)
...
+            if (leftType is PointerType && rightType is PointerType)
+            {
+                if (Operator == BinaryOperator.Add)
+                {
+                    throw new CompilationException("Operator '+': cannot add two pointers");
+                }
+
+                if (Operator == BinaryOperator.Subtract)
+                {
+                    throw new WipException(260, "Pointer substraction not implemented.");
```
ForNeVeR, line 79:
```suggestion
                    throw new WipException(260, "Pointer subtraction not implemented.");

```

Cesium.CodeGen/Ir/Types/CTypeSystem.cs
```diff
@@ -16,4 +16,6 @@ internal class CTypeSystem
     public IType CharPtr { get; } = new PrimitiveType(PrimitiveTypeKind.Char).MakePointerType();
     public IType Float { get; } = new PrimitiveType(PrimitiveTypeKind.Float);
     public IType Double { get; } = new PrimitiveType(PrimitiveTypeKind.Double);
+
+    public int GetTypeSize(IType type) => type.SizeInBytes;
```
ForNeVeR, line 20: Why do we need this? So now there's two ways of achieving the same.

Cesium.IntegrationTests/pointers/pointers_substraction.ignore.c
```diff
@@ -0,0 +1,7 @@
+int main(int argc, char* argv[])
+{
+    int a[10];
+    int* b = &a[1];
+    int* c = &a[8];
+    return 6 * (c - b);
+}
```
ForNeVeR, line 7: Please rename this file to `pointers_subtraction.ignore.c`.

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
kant2002: I'm converting expression which would be added to pointer. nit: we don't have `nint`, probably closer alternative in C would be `ptrdiff_t`. Internally it would be `nint`, but we express conversion in term of C language here.
This is probably place where I would like some changes in the expression tree. for example have `PointerAdditionOperatorExpression` which will contains all these conversions encoded (including implicit conversion as we have here), so we would not complicate simple math operations.

Cesium.CodeGen/Ir/Types/CTypeSystem.cs
kant2002: I would like to get rid of `SizeInBytes` as away to get type size. My opinion that Type knows too much about thinks, but maybe you are right here. When I prepare answer I probably found answer to #132

Cesium.IntegrationTests/pointers/pointers_substraction.ignore.c
kant2002: Eventually I fix my bad habits

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
ForNeVeR: I believe that we should introduce `nint` to our type system. Yes, it isn't a real primitive C type, but it implicitly exists for this exact purpose (to express types of expressions added/subtracted to/from pointers).

Cesium.CodeGen/Ir/Types/CTypeSystem.cs
ForNeVeR: I don't think that's an answer to #132, unfortunately, but that's an interesting thought anyway.

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
kant2002: I beleive this is `intptr_t`/`ptrdiff_t` which is optional typedef. Question does not `nint` or `nuint` ?  It seems that we use `nuint` based on this https://github.com/ForNeVeR/Cesium/pull/262/files#diff-e147e0aabdd129cb04f0d5838b89426531b2226cc6fcd5397a434aed085366d4R5

ForNeVeR: I'd use `nint` for now, perhaps. Though considering how .NET handles overflows, I'm not sure this particular choice will matter much.

The choice of type size will matter, but signed vs unsigned in this context – _probably_ not.

kant2002:
@ForNeVeR ready to review

ForNeVeR (approved):
Thanks!

PR #262, opened 2022-08-26
Fix issue when second variable was not registered

kant2002:


ForNeVeR:
Could you please take a look at the tests?

ForNeVeR (approved):
Thanks!

PR #265, opened 2022-08-27
Allow produce compilation output in same folder as Cesium

kant2002:
Before that compilation carash during copy of Cesium.Runtime.dll
Also copy not Cesium.Runtime which sit near compiler, but one which specified in command line

### DIFF
--- a/Cesium.Compiler/Compilation.cs
+++ b/Cesium.Compiler/Compilation.cs
@@ -100,7 +100,7 @@ internal static class Compilation
 
         // Prevent copying of the Cesium.Runtime if compile in same directory as compiler.
         var outputExecutablePathFullPath = GetNormalizedFullPath(outputExecutablePath);
-        var compilerPathFullPath = GetNormalizedFullPath(AppContext.BaseDirectory);
+        var compilerPathFullPath = GetNormalizedFullPath(Path.GetDirectoryName(compilerRuntimeDll) ?? Environment.CurrentDirectory);
         if (outputExecutablePathFullPath != compilerPathFullPath)
         {
             File.Copy(compilerRuntimeDll, applicationRuntime, true);
--- a/Cesium.Compiler/Compilation.cs
+++ b/Cesium.Compiler/Compilation.cs
@@ -27,7 +27,7 @@ internal static class Compilation
             await GenerateCode(assemblyContext, inputFilePath);
         }
 
-        SaveAssembly(assemblyContext, compilationOptions.TargetRuntime.Kind, outputFilePath);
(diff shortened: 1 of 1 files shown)


Cesium.Compiler/Compilation.cs
```diff
@@ -88,16 +88,23 @@ private static async Task GenerateCode(AssemblyContext context, string inputFile
...
         var applicationRuntime = Path.Combine(outputExecutablePath, "Cesium.Runtime.dll");
-        File.Copy(compilerRuntime, applicationRuntime, true);
+
+        // Prevent copying of the Cesium.Runtime if compile in same directory as compiler.
+        var outputExecutablePathFullPath = GetNormalizedFullPath(outputExecutablePath);
+        var compilerPathFullPath = GetNormalizedFullPath(AppContext.BaseDirectory);
+        if (outputExecutablePathFullPath != compilerPathFullPath)
+        {
+            File.Copy(compilerRuntimeDll, applicationRuntime, true);
+        }
```
ForNeVeR, line 107: Is this the situation when we copy the file into itself? If so, then maybe it would be better to compare `compilerRuntimeDll` and `applicationRuntime`?

As a bonus, you won't need the separator tricks to compare them, since both will be full file paths.

ForNeVeR (changes requested):

Cesium.Compiler/Compilation.cs
kant2002: Command line produce
```
Cesium.Compiler.exe C:\Cesium\Cesium.IntegrationTests\hello.c -o test.exe --modulekind console
```
have following values
```
compilerRuntime: C:\Cesium\artifacts\bin\Cesium.Compiler\Debug\net6.0\Cesium.Runtime.dll
applicationRuntime: Cesium.Runtime.dll
```
When call `Path.GetFullPath("", AppContext.BaseDirectory)` it return `C:\Cesium\artifacts\bin\Cesium.Compiler\Debug\net6.0` but `Path.GetFullPath(compilerRuntime, AppContext.BaseDirectory)` return `C:\Cesium\artifacts\bin\Cesium.Compiler\Debug\net6.0\` so I think normalization is required

ForNeVeR: I still think it's better to compare `Path.GetFullPath(compilerRuntime)` and `Path.GetFullPath(applicationRuntime)`.

ForNeVeR: I mean, the invariant of taking one of the artifacts from `AppContext.BaseDirectory` is very fragile. To make the code more robust, it's better to directly protect from what's happening: from copying the file into itself.

ForNeVeR (approved):
Thanks!

PR #266, opened 2022-08-27
Output all build files to single folder

kant2002:
That gives nice property that someone can nuke all build artifacts just by deliting folder, but also allows easier testing Nuget packages, since all of them would be stored in same `artifacts\packages\$(Configuration)\Shipping` folder for example.

Also add ZIP distribution for the compiler. At this point this file can be safely published and distributed, and if someone have .NET 6 installed, can see the glory of current Cesium.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ obj/
 *.runtimeconfig.json
 *.received.txt
 *.user
+*.binlog
+artifacts/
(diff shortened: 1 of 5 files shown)


ForNeVeR:
I'll be honest. in this case, you're introducing complexity for (incomplete) publishing/deployment processes which aren't here, yet.

I lean towards _not merging_ this without explanation of some very good benefits for the current compiler developers (i.e. ourselves), or full testing chain which will show how to use these tasks.

Without testing on CI, this all is dead code with unclear invariants which have to be maintained via unclear (to me) means.

Also, I personally believe that this is a lot of MSBuild code which should really be done in scope of the CI/CD process (maybe in form of scripts) and not actual build.

Issue #243, opened 2022-08-17
Support for `float` type


ForNeVeR:
As of #189, we have support for `double`, but still none for `float`. We should support that.

A good start is to look for `// TODO[#243]` in the code.

ForNeVeR:
This turns out to be harder than I thought.

In the C source, all (most?) of the constants are of `double` type, while there may be some target-typing involved (so `float x = 1.0` should still work).

Maybe more thought should be put into target-typing of expressions?

PR #267, opened 2022-08-28
Add typecast expression

### LINKED ISSUE #246: Insert implicit casts during IR building
In https://github.com/ForNeVeR/Cesium/pull/244 commented out `CodeGenOperatorTests::AddIntToChar` test. 
It should be resurrected by adding type conversion IR nodes.

kant2002:
- Make typecast explicit operation in the IR.
- Due to fact that `IDeclarationScope` is inaccessible during Lower call. I put explicit typecasts during emit phase. That require subsequet changes in `IExpression` interface and possible in the `ToIntermediate` as well.
- `BinaryOperatorExpression.EmitConversion` consult now `CTypeSystem`, I plan to remove that method altogether and replace it by insertion of corresponding `TypeCastExpression` when needed.
- Changes in `PrimitiveTypes` test is artifact of typecast insertion. Once constant evaluation would start appearing, I expect changes in these area again. Same for `FloatConstTest2`, it will emit ldc.r4 becasue `(float)1.5` would be folded directly to `1.5f`

I did not close #246 with that PR because I need to invest in more tests for conversions, and rework of `ArithmeticBinaryOperatorExpression` to insert `TypeCastExpression`. Basically I expect during lowering following conversion `IntConstant(1) + CharConstant('c')` => `IntegerConstant(1) + IntegerConstant(99)`, but introduction of `TypeCastExpression` is allow to work on that issue.

Closes: #243

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
@@ -80,6 +80,8 @@ int main()
     [Fact] public Task Parameter5Get() => DoTest("int foo(int a, int b, int c, int d, int e){ return e + 1; }");
     [Fact] public Task CharConstTest() => DoTest("int main() { char x = '\\t'; return 42; }");
     [Fact] public Task DoubleConstTest() => DoTest("int main() { double x = 1.5; return 42; }");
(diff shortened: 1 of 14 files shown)


ForNeVeR (approved):
Thanks!

PR #268, opened 2022-08-28
Explicitly inject type cast for addition/and comparison operators

kant2002:
- Move type conversion emit to TypeCast logic.
- All typecast insertion logic now happens during Lower call


Cesium.CodeGen/Ir/Expressions/IdentifierExpression.cs
```diff
@@ -25,7 +25,12 @@ public IdentifierExpression(Ast.IdentifierExpression expression)
         Identifier = expression.Identifier;
```
ForNeVeR, line 25: Let's make this to call `this(expression.Identifier)`, then?

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

Issue #201, opened 2022-08-14
Improve the compilation pipeline (introduce stages)


ForNeVeR:
For now, we have somewhat messed up pipeline: the IR nodes do all by themselves. Type checking, member resolve, lowering, codegen: we have everything everywhere. Right now, only parsing is properly isolated from everything else.

We should introduce more formal stages of compilation and _maybe_ even create several different layers of IR? For example, we have certain constructs that have to be lowered (such as `+=`-styled operators). We may get rid of these constructs in the "lowered IR layer" and thus decrease the amount of type checks required in the codegen.

I don't know how should it be organized, though. Just start from separating the IR layers, and everything else will click and fit in place?

Thoughts?

cc @kant2002, @impworks

impworks:
I would suggest having the following compilation stages:

1. Parsing: turns raw input into AST. The AST at this stage may be ambiguous.
2. Resolution and lowering: performs all type checks, resolves identifiers, etc. This level knows nothing about .NET type system and only operates on types defined in the C spec. Depending on the semantics, each AST node is mapped to some CST node which is unambiguous and desugared (e.g. implicit casts are inserted, shorthand assignments expanded, etc.)
3. Emit: CST nodes are emitted to Mono.Cecil structures and MSIL opcodes. C types are mapped to .NET types.

Not sure about the optimizations thought. It might be wise to add a special step between 2 and 3, which will fold constants, eliminate dead code, unroll loops etc. - or it might be simpler to just do that kind of processing directly in the "emit" stage.

The visitor pattern would not probably help with this: it's a good solution for the "expression problem" when the set of nodes is finite, and the set of operations is not (e.g. can be extended by the user). Since we do not intend to provide any kind of API for compiler plugins, a simple OOP-based solution will work just fine.

kant2002:
I agree with @impworks There some points also very important for me
- This level knows nothing about .NET type system and only operates on types defined in the C spec
- "The visitor pattern would not probably help with this". I do think that visitor pattern is overkill at least for now, until we figure out problem space better.

Other parts which I think important in regarding to CST
1. [Basic blocks for CFG](https://www.geeksforgeeks.org/software-engineering-control-flow-graph-cfg/). If we split execution graph on Basic blocks that simplify our branching structure. Only BB would know about branching, other parts of CST would be linear. That's usually first part of transformation.
[...]

PR #270, opened 2022-09-01
Split emit and declaration scopes

kant2002:
- All IL related stuff should be in Emit scope
- Declaration scope should operate only on IR model.

Food for thoughts:
Inheritance relationship between IEmitScope and IDeclaration scope probably not needed. 
IValue is leaky abstraction, since it operates at emit level, but not at compile time level. That triggers compilation error much later in the compilation process, at the emit level, but it should be at Lower call or earlier. There should be one more step where types and identifiers would be resolved as per suggestion. Reason why this is needed - NamedType which has only identifier, but instead of should perform lookup on type in the global alias table and not inside assembly.

Moving towards: #201

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/Values/LValueLocalVariable.cs
+++ b/Cesium.CodeGen/Ir/Expressions/Values/LValueLocalVariable.cs
@@ -7,18 +7,20 @@ namespace Cesium.CodeGen.Ir.Expressions.Values;
 
 internal class LValueLocalVariable : ILValue
 {
-    private readonly VariableDefinition _definition;
     private readonly IType _variableType;
+    private readonly string _name;
+    private VariableDefinition? _definition;
 
-    public LValueLocalVariable(IType variableType, VariableDefinition definition)
+    public LValueLocalVariable(IType variableType, string name)
     {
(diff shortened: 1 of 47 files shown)


Cesium.CodeGen/Ir/Expressions/Values/LValueLocalVariable.cs
```diff
@@ -29,23 +31,36 @@ public void EmitGetValue(IDeclarationScope scope)
...
+    private VariableDefinition GetVariableDefinition(IEmitScope scope)
+    {
+        if (_definition != null)
+        {
+            return _definition;
+        }
+
+        _definition = scope.ResolveVariable(_name);
+        return _definition;
+    }
```
ForNeVeR, line 65:
```suggestion
    private VariableDefinition GetVariableDefinition(IEmitScope scope) =>
        _definition ??= scope.ResolveVariable(_name);
```

ForNeVeR (changes requested):

ForNeVeR:
Other than the mentioned change, LGTM!

ForNeVeR (approved):
Thanks!

PR #271, opened 2022-09-09
Split emit and declaration scopes

kant2002:
Have to create separate expressions for IValue elements.

### DIFF
--- a/Cesium.CodeGen/Contexts/ForScope.cs
+++ b/Cesium.CodeGen/Contexts/ForScope.cs
@@ -7,22 +7,22 @@ using Mono.Cecil.Cil;
 
 namespace Cesium.CodeGen.Contexts;
 
-internal record ForScope(IEmitScope Parent) : IEmitScope
+internal record ForScope(IEmitScope Parent) : IEmitScope, IDeclarationScope
 {
     public AssemblyContext AssemblyContext => Parent.AssemblyContext;
     public ModuleDefinition Module => Parent.Module;
     public CTypeSystem CTypeSystem => Parent.CTypeSystem;
-    public IReadOnlyDictionary<string, FunctionInfo> Functions => Parent.Functions;
+    public IReadOnlyDictionary<string, FunctionInfo> Functions => ((IDeclarationScope)Parent).Functions;
     public TranslationUnitContext Context => Parent.Context;
     public MethodDefinition Method => Parent.Method;
-    public IReadOnlyDictionary<string, IType> Variables => Parent.Variables; // no declarations for `for` now, so pass parent variables
-    public IReadOnlyDictionary<string, IType> GlobalFields => Parent.GlobalFields;
+    public IReadOnlyDictionary<string, IType> Variables => ((IDeclarationScope)Parent).Variables; // no declarations for `for` now, so pass parent variables
+    public IReadOnlyDictionary<string, IType> GlobalFields => ((IDeclarationScope)Parent).GlobalFields;
     public void AddVariable(string identifier, IType variable) =>
         throw new WipException(205, "Variable addition into a for loop scope is not implemented, yet.");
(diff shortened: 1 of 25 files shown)


Cesium.CodeGen/Contexts/ForScope.cs
```diff
@@ -7,22 +7,22 @@
...
     public ModuleDefinition Module => Parent.Module;
     public CTypeSystem CTypeSystem => Parent.CTypeSystem;
-    public IReadOnlyDictionary<string, FunctionInfo> Functions => Parent.Functions;
+    public IReadOnlyDictionary<string, FunctionInfo> Functions => ((IDeclarationScope)Parent).Functions;
     public TranslationUnitContext Context => Parent.Context;
     public MethodDefinition Method => Parent.Method;
-    public IReadOnlyDictionary<string, IType> Variables => Parent.Variables; // no declarations for `for` now, so pass parent variables
-    public IReadOnlyDictionary<string, IType> GlobalFields => Parent.GlobalFields;
+    public IReadOnlyDictionary<string, IType> Variables => ((IDeclarationScope)Parent).Variables; // no declarations for `for` now, so pass parent variables
+    public IReadOnlyDictionary<string, IType> GlobalFields => ((IDeclarationScope)Parent).GlobalFields;
```
ForNeVeR, line 19: The casts here are alarming. Should this block have two differently-typed parents now?

Cesium.CodeGen/Ir/TopLevel/TopLevelDeclaration.cs
```diff
@@ -88,7 +88,7 @@ public void EmitTo(TranslationUnitContext context)
         {
             var field = context.AssemblyContext.ResolveGlobalField(name, context);
             var globalInitializerScope = context.GetInitializerScope();
-            initializer.Lower(globalInitializerScope).EmitTo(globalInitializerScope);
+            initializer.Lower((IDeclarationScope)globalInitializerScope).EmitTo(globalInitializerScope);
```
ForNeVeR, line 91: I suggest getting rid of this cast by making `context.GetInitializerScope()` to return a `GlobalConstructorScope`. It's internal, anyway, so not a big deal.

ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/ForScope.cs
kant2002: I try to limit amount of changes which I put in single PR. for now, scopes are should implement two interfaces. And I can pinky swear that I get rid of that. but that's potentially additional drag. At least now all developers cannot cross boundaries of syntax analysis and emit.

If I pass tow scopes for `ForScope` I just move that cast to [different location](https://github.com/ForNeVeR/Cesium/blob/3e07a3bf96f3fe24596af45d75fbcf0a741d3d54/Cesium.CodeGen/Ir/BlockItems/ForStatement.cs#L50). I do not see point to have `ForEmitScope` at all, but current implementation is relying on that, so I have to keep it for sure. but `ForDeclarationScope` is pointless since it was never actually used. I prefer leave this, knowing this is very problematic place which should be solved.

kant2002: Also, this chain of refactoring made me realize that maybe `IEmitScope` and `TranslationUnitContext` is same thing. Not yet sure, but both things look pretty close in spirit to each other.

ForNeVeR: > If I pass tow scopes for ForScope I just move that cast to [different location](https://github.com/ForNeVeR/Cesium/blob/3e07a3bf96f3fe24596af45d75fbcf0a741d3d54/Cesium.CodeGen/Ir/BlockItems/ForStatement.cs#L50).

Ok, this is fair. Let it be, then.

ForNeVeR (approved):
Thanks!

Issue #232, opened 2022-08-14
Complex type array support


ForNeVeR:
Look for the number `232` in the code to find clues to implement this feature.

kant2002:
Sample test case below, and can be found in `array\array_struct.c` file
```
typedef struct { int x; } foo;

int main(int argc, char *argv[])
{
    foo a[10];
    a[2 - 1].x = 42;
    return a[1].x;
}
```

Issue #214, opened 2022-08-14
Support typedef at block level


ForNeVeR:
Currently, we don't support this:

```
void foo() {
  typedef int xxx;
  xxx x;
} 
```

We should support this.

Look for the number `xxx` in the code to find clues to implement this feature.

PR #273, opened 2022-09-13
Resolve all types into internal model

kant2002:
That allows implementing of the #232, I think that's only thing which prevents #252 from finishing and overall fills the gap at Lower level where we do not know types.


kant2002:
Ooh, and #214 should be doable

Cesium.CodeGen/Contexts/IDeclarationScope.cs
```diff
@@ -11,5 +11,7 @@ internal interface IDeclarationScope
     IReadOnlyDictionary<string, IType> Variables { get; }
     IReadOnlyDictionary<string, IType> GlobalFields { get; }
     void AddVariable(string identifier, IType variable);
+    IType DisambiguateType(IType type);
```
ForNeVeR, line 14: Could you please explain what this method is supposed to do? Or maybe even add the xmldoc here.

ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/IDeclarationScope.cs
kant2002: This method find real C type based on the name of type or return existing if all type information already captured. 

I believe this should be separate stage of approximately following pipeline
1. Registration of declarations
2. Type resolution (this method and maybe something else)
3. Code transformation(Lower)
4. Emit 

I also suspect that we choose Lower as method name incorrectly as I think in compiler development this means very specific stage.

ForNeVeR: So, this is actually `Resolve`?

To me, the signature looks pretty strange: why does it take an `IType`? If I want a type, then presumably I shouldn't have one?

kant2002: Technically we have only NamedType which is half baked, but it can be anywhere in the concrete ITypes

ForNeVeR: What if we limit this to only accept `NamedType`? Or even just to accept the name? This would force the reader to figure out what to do with the rest of the type.

kant2002: We need to resolve StructType in which NamedType are type of the field for example. Or if NamedType is base type for array. 

This function is almost type resolution phase which I cannot do now. Currently I at least two small refactoring away from talking with you about clear stages. First step is conversion of ScopedIdentifierDeclaration into IBlockDeclaration. Then do the same for FunctionDeclaration and probably some wrap up of what’s left. After that we can work on separate phases. That’s mundane but not obvious work with delayed benefits. At least after hiding stage in the function I can unblock one potential contributor and couple new work items. 

But if you have ideas what I can do better, I willing to change something or go into other direction.

ForNeVeR: Okay. After re-reading the imlementation, I believe it is now correct and no, we cannot limit it to `NamedType`.

I suggest naming it either `FullyResolve`, or `RecursivelyResolve`, or `DeepResolve`, because that's what it really does in compiler terms: it does name resolve for the type itself and for all its members recursively.

Another naming variant, if you still insist on "resolve" being something else, then we can call it `Materialize` or `Instantiate`.

And we should add an explaining comment to this method. I suggest something like this:
```
/// Recursively resolve the passed type and all its members, replacing `NamedType` in any points with their actual instantiations in the current context.
///
/// Throws a <see cref="CompilationException"/> if it's not possible to resolve some of the types.
```

ForNeVeR (changes requested):
Except for that naming debate, everything else is LGTM.

Cesium.CodeGen/Contexts/IDeclarationScope.cs
kant2002: I would vote for `ResolveType` since we cannot resolve type "not fully". There circumstances where our code can work on unresolved types, like some usages of pointers like `foo*` but that's mean that we do not resolve type yet.

ForNeVeR: Alright, that will work.

ForNeVeR (approved):
Thanks!

Issue #276, opened 2022-09-18
Offsetof implementation


kant2002:
Standard requries stddef .h to be defined. 
Example implementation
```c
#define offsetof(s,m) ((size_t)&(((s*)0)->m))
```
That assumes that pointer to structure is stored in the linear memory. With current Cesium model I do not see obvious way how to implement that without resorting to intrinsic. And even then, I cannot see what intrinsic should we provide to implement this one feature

ForNeVeR:
From my experiments: this doesn't seem to work in C#, even if it only uses `ldflda`:
```
struct Something
{
    public int memberA, memberB;
}

public class C {
    public unsafe static void Main() {
        var nullptr = ((Something*)null); 
        var offset = (byte*)&nullptr->memberB;
        Console.WriteLine((int)offset);
    }
}
```
It throws a `NullReferenceException` when trying to get `&nullptr->memberB`.

ForNeVeR:
Some thoughts:
1. Yep, seems like this _has to_ be an intrinsic.
2. If the type information is available (and it necessarily is), then we should be able to implement that intrinsic easily*

* Obviously it won't be easy, since we'll have to duplicate the runtime code that calculates object layout. But still doable, since AFAIK the layout for `StructLayoutKind.Sequential` is standardized.

PR #277, opened 2022-09-18
Add declarations for stddef.h

kant2002:
- exposes nint type. That's Cesium specific type which maps to `System.IntPtr`. That allow implement `ptrdiff_t`, `intptr_t` and `uintptr_t` in unified manner.
- `offsetof` does not implemented. See #276

### DIFF
--- a/Cesium.CodeGen/Ir/Declarations/LocalDeclarationInfo.cs
+++ b/Cesium.CodeGen/Ir/Declarations/LocalDeclarationInfo.cs
@@ -303,6 +303,7 @@ internal record LocalDeclarationInfo(
                 ("long", null, null, null) => PrimitiveTypeKind.Long,
                 ("float", null, null, null) => PrimitiveTypeKind.Float,
                 ("double", null, null, null) => PrimitiveTypeKind.Double,
+                ("__nint", null, null, null) => PrimitiveTypeKind.NativeInt,
                 _ => throw new WipException(
                     224,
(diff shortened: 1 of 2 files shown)


Cesium.CodeGen/Ir/Declarations/LocalDeclarationInfo.cs
```diff
@@ -303,6 +303,7 @@ private static (IType, string? Identifier) ProcessDirectDeclarator(IDirectDeclar
                 ("long", null, null, null) => PrimitiveTypeKind.Long,
                 ("float", null, null, null) => PrimitiveTypeKind.Float,
                 ("double", null, null, null) => PrimitiveTypeKind.Double,
+                ("nint", null, null, null) => PrimitiveTypeKind.NativeInt,
```
ForNeVeR, line 306: What would you say about calling that either `__nint` or `_Nint`? more in line with how the standard works.

(and then, we _could_ introduce a Cesium-only file that would add a `typedef nint __nint`, if really required)

kant2002: I dont belive we need both `__nint` and `nint`. I only want expose that implementation detail, so we can model `ptrdiff_t`, `intptr_t` and `uintptr_t` properly. I do not even want to document that type.

ForNeVeR (approved):
Thanks!

PR #278, opened 2022-09-18
Add test cases for multidimensional arrays

kant2002:
Just one more test and extract test to separate class, so we can at least group related functionality


ForNeVeR (approved):
Thanks!

Issue #196, opened 2022-08-14
Functions with varargs support


kant2002:
We need varargs support, since almost any simple tutorial use `printf`. Without that support we cannot go to public probably, otherwise Cesium always look as toy project.

ForNeVeR:
I agree on that point. Any ideas how such a function could work? The problem is that sometimes it's hard to distinguish a vararg function from the call-site point of view: it may look (and even be declared?) as a normal function.

Could we reuse the existing facilities of vararg functions in the CLI? Do these facilities even exist?

kant2002:
https://github.com/dotnet/runtime/issues/10478 that's not possible today, even if it would be possible in the future. So plan would be to mimic this. I think we can pass variadic arguments as `object[]` and that's it. That's will make implementing va_start ambiguous when used incorrectly, but I think that fine. That unlock interop scenarios, which is what's important at this stage. If variadic function would be declared as regular one, and defined as variadic that's we should be able resolve at "link" time, and that's pretty advanced and arcane scenario. I do not think we can afford think too much about it, give that we do not know space very well

ForNeVeR:
> I think we can pass variadic arguments as `object[]` and that's it.

To be honest, I'm not sure we can put a pointer into an `object[]`. Is pointer even an object? 🤔 

Maybe some workaround using stack arrays or linked lists will be required.

kant2002:
I do find some interesting sample which definitely will complicate our life.
```
#include <stdio.h>
#include <stdarg.h>

void function1(char* s, ...)
{
    va_list ap;
    int tmp;

    va_start(ap, s);
    tmp = va_arg(ap, int);
    printf(s, tmp);
    va_end(ap);
}

void function2(char* s, int d)
{
    printf(s, d);
}

typedef void (*functionX_t)(char*, int);
typedef void (*functionY_t)(char*, ...);

int main(int argc, char* argv[])
{
    int x = 42;

    /* swap! */
    functionX_t functionX = (functionX_t) function1;
    functionY_t functionY = (functionY_t) function2;

    function1("%d\n", x);
    function2("%d\n", x);
    functionX("%d\n", x);
    functionY("%d\n", x);

    return 0;
}
```

ForNeVeR:
What does the standard says about this? I'm not sure these signatures are technically compatible.

BadRyuner:
[...]

PR #279, opened 2022-09-25
Primitive support for varargs

kant2002:
For now only char* arguments supported. Cannot do much with integers, since during emit phase I have lost information about expression type, and I cannot imagin what kind of IR node I need to preserve that, and at the same time do not overcomplicate things. Related to #196

### DIFF
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -45,6 +45,8 @@ public unsafe static class StdIoFunctions
                     Console.Write(Unmarshal((byte*)(IntPtr)varargs[consumedArgs]));
                     consumedArgs++;
                     break;
+                default:
+                    throw new FormatException($"Format specifier {formatSpecifier} is not supported");
             }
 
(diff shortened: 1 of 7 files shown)


Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -41,4 +23,59 @@ public static void PutS(byte* str)
...
+        {
+            Console.Write(formatString.Substring(currentPosition, formatStartPosition - currentPosition));
+            var formatSpecifier = formatString[formatStartPosition + 1];
+            switch (formatSpecifier)
+            {
+                case 's':
+                    Console.Write(Unmarshal((byte*)(IntPtr)varargs[consumedArgs]));
+                    consumedArgs++;
+                    break;
+            }
```
ForNeVeR, line 48: What do you think of adding `throw` for a default case?

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #280, opened 2022-09-26
Pass varargs via stack allocated buffer

kant2002:
When invoke function annotated as vararg, created local buffer for the variable arguments. For each argument currently allocated 8 bytes. This should be enough for most common usage.

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
+++ b/Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
@@ -39,10 +39,36 @@ internal class FunctionCallExpression : IExpression
         var functionName = _function.Identifier;
         var callee = scope.Functions.GetValueOrDefault(functionName)
                      ?? throw new CompilationException($"Function \"{functionName}\" was not found.");
+        int firstVarArgArgument = 0;
+        if (callee.Parameters is { } parameters)
+        {
+            if (parameters.IsVarArg)
+            {
+                firstVarArgArgument = parameters.Parameters.Count;
+            }
+        }
+
         return new FunctionCallExpression(
             _function,
             callee,
-            _arguments.Select(a => a.Lower(scope)).ToList());
+            _arguments.Select((a, index) =>
+            {
+                if (index >= firstVarArgArgument)
+                {
+                    var expressionType = a.GetExpressionType(scope);
+                    if (expressionType.Equals(scope.CTypeSystem.Float))
+                    {
+                        // Seems to be float always use float-point registers and as such we need to covert to double.
+                        return new TypeCastExpression(scope.CTypeSystem.Double, a.Lower(scope));
+                    }
+                    else
+                    {
+                        return a.Lower(scope);
+                    }
+                }
+
+                return a.Lower(scope);
+            }).ToList());
     }
 
     public void EmitTo(IEmitScope scope)
@@ -81,7 +107,6 @@ internal class FunctionCallExpression : IExpression
                 scope.AddInstruction(OpCodes.Ldc_I4, i * 8);
                 scope.AddInstruction(OpCodes.Add);
                 argument.EmitTo(scope);
-                scope.AddInstruction(OpCodes.Conv_I);
                 scope.AddInstruction(OpCodes.Stind_I);
             }
 
--- a/Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
+++ b/Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
@@ -75,6 +75,7 @@ internal class FunctionCallExpression : IExpression
         var varArgParametersCount = _arguments.Count - explicitParametersCount;
         if (_callee!.Parameters?.IsVarArg == true)
         {
+            // TODO: See https://github.com/ForNeVeR/Cesium/issues/285
             // Using sparse population of the parameters on the stack. 8 bytes should be enough for anybody.
             // Also we need perform localloc on empty stack, so we will use local variable to save vararg buffer to temporary variable.
             if (varArgParametersCount == 0)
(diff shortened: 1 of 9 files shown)


Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
```diff
@@ -39,36 +39,78 @@ public IExpression Lower(IDeclarationScope scope)
         var functionName = _function.Identifier;
         var callee = scope.Functions.GetValueOrDefault(functionName)
                      ?? throw new CompilationException($"Function \"{functionName}\" was not found.");
+        int firstVarArgArgument = 0;
+        if (callee.Parameters is { } parameters)
+        {
+            if (parameters.IsVarArg)
```
ForNeVeR, line 45: _(Minor)_
```suggestion
        if (callee.Parameters?.IsVarArg == true)
        {
```

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
```diff
@@ -39,36 +39,78 @@ public IExpression Lower(IDeclarationScope scope)
...
     }
 
     public void EmitTo(IEmitScope scope)
     {
+        VariableDefinition? varArgBuffer = null;
         var explicitParametersCount = _callee!.Parameters?.Parameters.Count ?? 0;
+        var varArgParametersCount = _arguments.Count - explicitParametersCount;
+        if (_callee!.Parameters?.IsVarArg == true)
+        {
+            // Using sparse population of the parameters on the stack. 8 bytes should be enough for anybody.
```
ForNeVeR, line 81: _(Minor)_ Are vararg struct parameters allowed? Because, you know, we can create a struct of any size.

(we can merge this as-is of course, that's more of a discussion matter and not a complaint)

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
```diff
@@ -39,36 +39,78 @@ public IExpression Lower(IDeclarationScope scope)
...
+            // Also we need perform localloc on empty stack, so we will use local variable to save vararg buffer to temporary variable.
+            if (varArgParametersCount == 0)
+            {
+                scope.AddInstruction(OpCodes.Ldc_I4_0);
```
ForNeVeR, line 85: What's the purpose of this `Ldc_I4_0`?

ForNeVeR (changes requested):
Please also add some codegen verification tests for this.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
kant2002: This is null.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
kant2002: Passing structs are possible. But I need to come up with lowering process which preserve information about expression type.  That simply need some work.

Cesium.IntegrationTests/stdlib/printf.c
```diff
@@ -4,16 +4,25 @@
...
+    printf("%lu\n", -1);
+
+    float floatValue = 5.99;
+    printf("%f\n", floatValue);
+    double doubleValue = 2.04;
+    printf("%f\n", doubleValue);
+
+    char myLetter = 'D';
+    printf("%c\n%c", myLetter, '1');
+    // We cannot validate this automatically, but at least we can uncomment and check
```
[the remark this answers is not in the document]
kant2002: This is printing pointer’s address. We can have test this as part of unit tests for Runtime library, but that require integration with native parts. Setting up cross platform automation would be interesting. CMake probably.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
kant2002: At this point I almost like to have all expressions to have special field/property `ExpressionType` which is set by some specific compiler phase.

Cesium.IntegrationTests/stdlib/printf.c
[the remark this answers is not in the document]
ForNeVeR: Ah okay, got it.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
ForNeVeR: Okay, let's at least add a TODO here. "8 bytes should be enough for anybody" sounds a little bit raw, especially if we know it isn't enough :)

ForNeVeR (changes requested):
No further complaints other than what we have already discussed.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
kant2002: Basically I borrow that from Sharplab. Ldnull works fine too and can change to it if you prefer, but seems to be both these methods are okay

ForNeVeR: I would prefer `ldnull`. Less questions that way.

Or, if this is some sort of optimization (say, `ldc_i4_0` takes less space in byte code), then let's have a helper method that chooses an appropriate form of `ldnull`-compatible instruction.

kant2002: Done. Anything else? Size is the same since no parameters passed to opcode

ForNeVeR (approved):
Thanks!

PR #282, opened 2022-09-26
Prepare for refactoring of declaration scopes

kant2002:
Declaration scopes inherently hierarchical, and we need to preserve that hierarchy. By exposing separate dictionaries we show too much of internal structure, even if all we need is resolve identifier, and know what kind of identifier it is. For example GlobalFields and Variables is essentially the same, except they are declared in different scopes. This is mostly not important until emit phase.

Rework for statement to not rely on `EndInstruction`
That makes emit scopes hierarchical, but we really want to have declaration scopes as hierarchical.

### DIFF
--- a/Cesium.CodeGen/Contexts/ForScope.cs
+++ b/Cesium.CodeGen/Contexts/ForScope.cs
@@ -39,7 +39,6 @@ internal record ForScope(IEmitScope Parent) : IEmitScope, IDeclarationScope
     public ParameterDefinition ResolveParameter(string name) => Parent.ResolveParameter(name);
     public ParameterInfo? GetParameterInfo(string name) => ((IDeclarationScope)Parent).GetParameterInfo(name);
 
-    public Instruction? EndInstruction { get; set; }
     /// <inheritdoc />
     public IType ResolveType(IType type) => Context.ResolveType(type);
     public void AddTypeDefinition(string identifier, IType type) => throw new AssertException("Not supported");
@@ -54,5 +53,19 @@ internal record ForScope(IEmitScope Parent) : IEmitScope, IDeclarationScope
     public Instruction ResolveLabel(string label)
     {
         return Parent.ResolveLabel(label);
+    }
+
+    /// <inheritdoc />
+    public void RegisterChildScope(IDeclarationScope childScope)
+    {
+
+    }
+
+    private string _breakLabel = Guid.NewGuid().ToString();
+
+    /// <inheritdoc />
+    public string? GetBreakLabel()
+    {
+        return _breakLabel;
     }
 }
--- a/Cesium.CodeGen/Contexts/ForScope.cs
+++ b/Cesium.CodeGen/Contexts/ForScope.cs
@@ -4,6 +4,8 @@ using Cesium.CodeGen.Ir.Types;
 using Cesium.Core;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Cesium.CodeGen.Contexts;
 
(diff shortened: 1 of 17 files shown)


Cesium.CodeGen/Contexts/IDeclarationScope.cs
```diff
@@ -1,16 +1,18 @@
...
+using System.Diagnostics.CodeAnalysis;
 
 namespace Cesium.CodeGen.Contexts;
 
 internal interface IDeclarationScope
 {
     CTypeSystem CTypeSystem { get; }
-    IReadOnlyDictionary<string, FunctionInfo> Functions { get; }
-    IReadOnlyDictionary<string, IType> Variables { get; }
+    bool TryGetFunctionInfo(string identifier, [NotNullWhen(true)] out FunctionInfo? functionInfo);
```
ForNeVeR, line 11: Let's make this to just return `FunctionInfo?`. No need to return `bool` and have this `out` parameter.

Cesium.CodeGen/Contexts/IDeclarationScope.cs
```diff
@@ -1,16 +1,18 @@
...
     IReadOnlyDictionary<string, IType> GlobalFields { get; }
     void AddVariable(string identifier, IType variable);
+    bool TryGetVariable(string identifier, [NotNullWhen(true)]out IType? type);
```
ForNeVeR, line 14: Same as above: let's return `IType?` and ditch `bool`.

Cesium.CodeGen/Contexts/ForScope.cs
```diff
@@ -1,43 +1,71 @@
...
+internal record ForScope(IEmitScope Parent) : IEmitScope, IDeclarationScope
+{
+    public AssemblyContext AssemblyContext => Parent.AssemblyContext;
+    public ModuleDefinition Module => Parent.Module;
+    public CTypeSystem CTypeSystem => Parent.CTypeSystem;
+    public bool TryGetFunctionInfo(string identifier, [NotNullWhen(true)] out FunctionInfo? functionInfo)
+        => ((IDeclarationScope)Parent).TryGetFunctionInfo(identifier, out functionInfo);
+    public TranslationUnitContext Context => Parent.Context;
+    public MethodDefinition Method => Parent.Method;
+    private readonly Dictionary<string, IType> _variables = new();
```
ForNeVeR, line 21: I can't find the code that fills this dictionary.

Cesium.CodeGen/Contexts/IDeclarationScope.cs
```diff
@@ -26,4 +28,16 @@ internal interface IDeclarationScope
     /// </summary>
     /// <param name="identifier">Label to add to the current scope.</param>
     void AddLabel(string identifier);
+
+    /// <summary>
+    // Registers child declaration scope.
+    /// </summary>
+    /// <param name="childScope">Child scope to add to the current one.</param>
+    void RegisterChildScope(IDeclarationScope childScope);
```
ForNeVeR, line 36: What's the use of this? It's never called.

Cesium.CodeGen/Contexts/IDeclarationScope.cs
```diff
@@ -26,4 +28,16 @@ internal interface IDeclarationScope
     /// </summary>
     /// <param name="identifier">Label to add to the current scope.</param>
     void AddLabel(string identifier);
+
+    /// <summary>
+    // Registers child declaration scope.
```
ForNeVeR, line 33:
```suggestion
    /// Registers child declaration scope.
```

Cesium.CodeGen/Ir/BlockItems/ForStatement.cs
```diff
@@ -28,26 +30,35 @@ public ForStatement(Ast.ForStatement statement)
...
     public IBlockItem Lower(IDeclarationScope scope)
-        => new ForStatement(
-            _initExpression?.Lower(scope),
-            _testExpression.Lower(scope),
-            _updateExpression?.Lower(scope),
-            _body.Lower(scope));
+    {
+        var forScope = new ForScope((IEmitScope)scope);
+        var breakLabel = forScope.GetBreakLabel();
+        scope.AddLabel(breakLabel!);
```
ForNeVeR, line 47: I think it's possible to declare `BreakLabel` as non-nullable in `ForScope`, to avoid `!` here.

ForNeVeR, line 47: I'm not entirely sure of side effects in `Lower`. The idea of `Lower` was that you can call it as much as you like (or even forget it). But now, `item.Lower().Lower()` can break the codegen.

Thoughts?

ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/ForScope.cs
kant2002: I hope transform ForScope into NestedScope in near future. Maybe I should not make code “for future use”

Cesium.CodeGen/Contexts/IDeclarationScope.cs
kant2002: Ooh yeah. I should remove this. This part I’m not even sure how it would be used.

Cesium.CodeGen/Ir/BlockItems/ForStatement.cs
kant2002: This is unfortunate hack. Proper handling of this requires separate compilation phase where I resolve labels and where I somehow control scope creation. Even if I have everything in place for multi phase compilation now, I do not sure how to properly implement such multiphase processing. I suspect that this clean architecture put performance pressure on us. Other concern over multiple calls to Lower which I have is that in that case we as developers loosing control of what’s going on. 

If you see any tricks, please let me know.

Cesium.CodeGen/Contexts/ForScope.cs
ForNeVeR: In this case, I think that's okay. But maybe we'll want to extract it into a base class.

Cesium.CodeGen/Ir/BlockItems/ForStatement.cs
ForNeVeR: Okay, I see that this issue has bigger scope. I'll note some places we'll need to replace in scope of #201.

ForNeVeR (approved):
Thanks.

PR #287, opened 2022-10-01
Support version command

### LINKED ISSUE #264: Version does not printed
Run command
```
Cesium.Compiler.exe --version
```
outputs
```
Cesium.Compiler 0.0.1
Copyright (C) 2022 Cesium.Compiler

  -o, --out        Required.

  --framework      (Default: Net)

  --modulekind

  --nologo         Suppress compiler banner message

  --namespace      Sets default namespace instead of "global"

  --globalclass    Sets default global class instead of "<Module>"

  --import         Provides path to assemblies which would be added as
                   references automatically into resulting executable.

  --corelib        Sets path to CoreLib assembly

  --runtime        Sets path to Cesium C Runtime assembly

  --help           Display this help screen.

  --version        Display version information.

  value pos. 0
```

Notice `value pos. 0` and also fact that version is ignored.

FrediKats:
close #264 

This is why it work, code of AutoBuild:

```csharp
public static HelpText AutoBuild<T>(ParserResult<T> parserResult, Func<HelpText, HelpText> onError, int maxDisplayWidth = DefaultMaximumLength)
{
    if (parserResult.Tag != ParserResultType.NotParsed)
        throw new ArgumentException("Excepting NotParsed<T> type.", "parserResult");

    var errors = ((NotParsed<T>)parserResult).Errors;

    if (errors.Any(e => e.Tag == ErrorType.VersionRequestedError))
        return new HelpText($"{HeadingInfo.Default}{Environment.NewLine}") { MaximumDisplayWidth = maxDisplayWidth }.AddPreOptionsLine(Environment.NewLine);
```

So they handle ErrorType.VersionRequestedError and use build-in logic for formatting version.

image

### DIFF
--- a/Cesium.Compiler/Main.cs
+++ b/Cesium.Compiler/Main.cs
@@ -42,16 +42,21 @@ return await parserResult.MapResult(async args =>
     },
     _ =>
     {
-        DisplayHelp(parserResult);
+        string helpText = PrepareHelpText(parserResult);
+        Console.WriteLine(helpText);
         return Task.FromResult(-1);
     });
 
-static void DisplayHelp<T>(ParserResult<T> result)
+static string PrepareHelpText<T>(ParserResult<T> result)
 {
+    if (result is NotParsed<T> notParsed && notParsed.Errors.IsVersion())
+        return HelpText.AutoBuild(result);
+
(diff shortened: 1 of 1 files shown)


ForNeVeR (approved):
Thanks!

PR #288, opened 2022-10-01
Add support for trivial #if directive

kant2002:
For now, only macro presence and equality/not-equality is working


ForNeVeR (changes requested):

ForNeVeR (approved):
Awesome, thank you!

PR #289, opened 2022-10-01
Disable build warnings

rstm-sf:
F. ex https://github.com/ForNeVeR/Cesium/actions/runs/3165184526/jobs/5154012086

### DIFF
--- a/Cesium.Compiler/Main.cs
+++ b/Cesium.Compiler/Main.cs
@@ -30,7 +30,7 @@ return await parserResult.MapResult(async args =>
 
         var cesiumRuntime = args.CesiumCRuntime ?? Path.Combine(AppContext.BaseDirectory, "Cesium.Runtime.dll");
         var defaultImportsAssembly = args.DefaultImportAssemblies ?? Array.Empty<string>();
-#pragma warning disable IL3000 Automatic discovery of corelib is fallback option, if tooling do not pass that parameter
+#pragma warning disable IL3000 // Automatic discovery of corelib is fallback option, if tooling do not pass that parameter
         var corelibAssembly = args.CoreLib ?? typeof(Math).Assembly.Location; // System.Runtime.dll
 #pragma warning restore IL3000
         var moduleKind = args.ModuleKind ?? Path.GetExtension(args.OutputFilePath).ToLowerInvariant() switch


Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -19,7 +19,9 @@ public static void PutS(byte* str)
         }
         catch (Exception) // TODO[#154]: Exception handling.
         {
+#pragma warning disable CS0219
```
kant2002, line 22: Personally, I would like to have this warning, so it will annoy me. That's guaranteed way to fix it in medium term

Cesium.IntegrationTests/Cesium.IntegrationTests.proj
```diff
@@ -1,4 +1,9 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="Current">
+
+    <PropertyGroup>
+        <NoWarn>NU1503</NoWarn>
```
kant2002, line 4: This one warning means that we have incomplete project file, and we should provide required target, even if would be empty one like other targets below.

Cesium.Compiler/Main.cs
```diff
@@ -30,7 +30,9 @@
 
         var cesiumRuntime = args.CesiumCRuntime ?? Path.Combine(AppContext.BaseDirectory, "Cesium.Runtime.dll");
         var defaultImportsAssembly = args.DefaultImportAssemblies ?? Array.Empty<string>();
+#pragma warning disable IL3000
```
kant2002, line 33: 👍

ForNeVeR: Please put an explanation in a comment here. Otherwise, it's hard to review and/or refactor if the code moves.

Cesium.IntegrationTests/Cesium.IntegrationTests.proj
ForNeVeR: I'd vote for adding a target framework (`net6.0`), even if unused. With a comment explaining the issue, of course.

Cesium.Runtime/StdIoFunctions.cs
ForNeVeR: Let's comment-out the `const int EOF` line instead.

ForNeVeR (changes requested):

Cesium.Compiler/Main.cs
kant2002:
```suggestion
#pragma warning disable IL3000 Automatic discovery of corelib is fallback option, if tooling do not pass that parameter
```

Cesium.IntegrationTests/Cesium.IntegrationTests.proj
kant2002: I would add `/bl` switch, look at the file and the provide necessary target. I think having just empty `_IsProjectRestoreSupported` target would be fine.

Cesium.Compiler/Main.cs
```diff
@@ -30,7 +30,9 @@
 
         var cesiumRuntime = args.CesiumCRuntime ?? Path.Combine(AppContext.BaseDirectory, "Cesium.Runtime.dll");
         var defaultImportsAssembly = args.DefaultImportAssemblies ?? Array.Empty<string>();
+#pragma warning disable IL3000 Automatic discovery of corelib is fallback option, if tooling do not pass that parameter
```
ForNeVeR, line 33: Hey, look, there are even more warnings now 😄 

Maybe it wants this?
```suggestion
#pragma warning disable IL3000 // Automatic discovery of corelib is fallback option, if tooling do not pass that parameter
```

ForNeVeR (changes requested):

PR #291, opened 2022-10-03
String escape sequence support

### LINKED ISSUE #235: Better string literal processing
Right now, we just strip the double quotes from a literal and then embed it into the binary.

We should process the escape sequences at least.

Look for the number `235` in the code to find clues to implement this feature.

BaLiKfromUA:
Hey folks!

I am trying to solve issue #235 but have two questions/problems:
- Did I choose the right direction to implement non-simple escape sequences or the solution with `Regex.Replace` is not what you expect to have?
- Two cases below are not interpreted by C (replaced by �) but interpreted by C# string/encoding utils. What did I miss?:(
```c
printf("\xF0"); // ð
printf("\xFF");// ÿ
```

Overall, I was trying to follow [this link](https://en.cppreference.com/w/c/language/character_constant) and fix [this integration test](https://github.com/ForNeVeR/Cesium/pull/275).

Closes #235.

### DIFF
--- a/Cesium.Parser/TokenExtensions.cs
+++ b/Cesium.Parser/TokenExtensions.cs
@@ -13,19 +13,58 @@ public static class TokenExtensions
         if (token.Kind != CTokenType.StringLiteral)
             throw new ParseException($"Non-string literal token: {token.Kind} {token.Text}");
 
+        var result = token.Text.Trim('"');
+
         // simple escape sequences
-        var result = token.Text.Trim('"')
-            .Replace("\\'", "\'")
-            .Replace("\\\"", "\"")
-            .Replace("\\?", "?")
-            .Replace("\\\\", "\\")
-            .Replace("\\a", "\a")
-            .Replace("\\b", "\b")
-            .Replace("\\f", "\f")
-            .Replace("\\n", "\n")
-            .Replace("\\r", "\r")
-            .Replace("\\t", "\t")
-            .Replace("\\v", "\v");
+        var builder = new StringBuilder(result.Length);
+        for (int i = 0; i < result.Length; ++i)
+        {
+            if (result.ElementAt(i) == '\\' && i < result.Length - 1)
+            {
+                switch (result.ElementAt(i + 1))
+                {
+                    case '\'':
+                    case '\"':
+                    case '?':
+                    case '\\':
+                        builder.Append(result.ElementAt(i + 1));
+                        break;
+                    case 'a':
+                        builder.Append('\a');
+                        break;
+                    case 'b':
+                        builder.Append('\b');
+                        break;
+                    case 'f':
+                        builder.Append('\f');
+                        break;
+                    case 'n':
+                        builder.Append('\n');
+                        break;
+                    case 'r':
+                        builder.Append('\r');
+                        break;
+                    case 't':
+                        builder.Append('\t');
+                        break;
+                    case 'v':
+                        builder.Append('\v');
+                        break;
+                    default:
+                        // todo: maybe smarter handling of this edge case with errors/warnings
+                        builder.Append("\\");
+                        --i; // don't skip next
+                        break;
+                }
+
+                ++i; // skip next
+            }
+            else
+            {
+                builder.Append(result.ElementAt(i));
+            }
+        }
+        result = builder.ToString();
 
         // numeric escape sequences
         result = Regex.Replace(result, @"\\([0-7]{1,3})", m =>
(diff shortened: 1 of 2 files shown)


kant2002:
Is it possible that you have � only because your terminal settings do not display these characters correctly? Other option is that we do not support locales properly and because of this you see differences in the output. For example, from terminal
```
Output for C:\Cesium\Cesium.IntegrationTests\strings\escape_sequence.c differs between native- and Cesium-compiled programs.
cl.exe (C:\Cesium\Cesium.IntegrationTests\strings\escape_sequence.c):
'"?\



?(???


Cesium (C:\Cesium\Cesium.IntegrationTests\strings\escape_sequence.c):
'"?\



?(∀🍌

strings\escape_sequence.c: failed.
```
Also I do have following warning from CL.exe
```
C:\Cesium\Cesium.IntegrationTests\strings\escape_sequence.c(19): warning C4566: character represented by universal-character-name '\u2200' cannot be represented in the current code page (1251)
C:\Cesium\Cesium.IntegrationTests\strings\escape_sequence.c(20): warning C4566: character represented by universal-character-name '\U0001F34C' cannot be represented in the current code page (1251)
```

`Regex.Replace` perfectly fine for me at this stage, since I do not see something this explicit.

Cesium.Parser/TokenExtensions.cs
```diff
@@ -11,7 +13,34 @@ public static string UnwrapStringLiteral(this IToken<CTokenType> token)
...
+
+        result = Regex.Replace(result, @"\\[xX]([0-9a-fA-F]{2})", m =>
+            char.ConvertFromUtf32(Convert.ToInt32("0x" + m.Groups[1].Value, 16)));
+
+        //universal character names
+        result = Regex.Replace(result, @"\\u([0-9a-fA-F]{4})", m =>
+            char.ConvertFromUtf32(Convert.ToInt32("0x" + m.Groups[1].Value, 16)));
+
+        result = Regex.Replace(result, @"\\U([0-9a-fA-F]{8})", m =>
+            char.ConvertFromUtf32(Convert.ToInt32("0x" + m.Groups[1].Value, 16)));
```
ForNeVeR, line 42: The general approach looks pretty okay, but this code creates a lot of temporary strings. Could we somehow improve that?

Say, for chained `Replace` you could use a `StringBuilder`, but then again: it is not ultra efficient and may be done better (with a single pass over the initial string).

When you develop a C compiler, it's sometimes hard to decide on what level of optimization you should stop.

ForNeVeR (changes requested):
I'd say let's improve this slightly (reduce allocations by chained `Replace`, but keep several `Regex` instances as is), and create a new issue about optimizing the hell out of this.

@kant2002, what do you think?

ForNeVeR:
Regarding your issue with Unicode output: I think that other compilers are unable to properly handle these strings, and this is not a problem of Cesium or your implementation. We may revisit this later when implementing locales (ideally, we should be able to make other compilers to output Unicode properly).

ForNeVeR:
And we should have green tests, of course. If this is caused by `\u` on Windows, then let's just disable the problematic cases for now.

Cesium.Parser/TokenExtensions.cs
kant2002: I’m fine with original approach. If @BaLiKfromUA want make this more efficient we can allocate `StringBuilder` with capacity same as original string and walk string in one pass replacing characters as needed. If that’s too complicated let us know. 
also make tests pass. Comment problematic cases as needed, but please do not remove.

BaLiKfromUA: I can implement it without any problems I guess. Initially, I just followed the original style + was trying to solve the task first.

kant2002: If you would like to join Russian speaking community you can go https://t.me/cilchat where we hanging out. If not, we can obviously communicate here.

BaLiKfromUA: I pushed my interpretation of your idea, hope it fits

BaLiKfromUA:
> And we should have green tests, of course. If this is caused by `\u` on Windows, then let's just disable the problematic cases for now.

Caused by Unicode, commented as mentioned

Cesium.Parser/TokenExtensions.cs
ForNeVeR: Yeah, it very much does.

ForNeVeR (approved):
Thanks!

PR #293, opened 2022-10-04
Add support for stdbool.h

### LINKED ISSUE #292: Implementing stdbool.h
I believe implementing this header would be good first issue. Just read 7.18 on page 211 and implement what's there

reima:
Closes #292.

I was thinking about adding an include guard, but decided against it, as currently no other stdlib header has one. Include guards should probably be added at some point to conform to 7.1.2 (4).

### DIFF
--- a/Cesium.IntegrationTests/stdlib/stdbool.c
+++ b/Cesium.IntegrationTests/stdlib/stdbool.c
@@ -1,0 +1,30 @@
+#include <stdbool.h>
+
+int main(int argc, char *argv[])
+{
+#ifndef bool
+    return 1;
+#endif
+
+#ifndef true
+    return 2;
+#endif
+
+#ifndef false
+    return 3;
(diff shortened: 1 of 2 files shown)


ForNeVeR:
> currently no other stdlib header has one

Mostly because some other headers were created before we added support for `#ifdef` :)

Cesium.IntegrationTests/stdlib/stdbool.c
```diff
@@ -0,0 +1,30 @@
...
+    if (true != 1) {
+        return 5;
+    }
+
+    if (false != 0) {
+        return 6;
+    }
+
+    return 42;
+}
```
ForNeVeR, line 30: Wow, that's a nice approach to testing! 👍

ForNeVeR (approved):
Thanks!

Issue #208, opened 2022-08-14
Implement support for all the C expressions


ForNeVeR:
After parsing all the expressions, we should implement full support for them in the code generator.

Look for the number `xxx` in the code to find clues to implement this feature.

Depends on:

- [ ] #207

Issue #207, opened 2022-08-14
Parse all the C expressions


ForNeVeR:
We should fully implement the following section of the C standard:
- [ ] 6.5 Expressions
  - [ ] 6.5.1 Primary expressions
  - [ ] 6.5.2 Postfix operators
  - [ ] 6.5.3 Unary operators
  - [ ] 6.5.4 Cast operators
  - [x] 6.5.5 Multiplicative operators
  - [x] 6.5.6 Additive operators
  - [ ] 6.5.7 Bitwise shift operators
  - [ ] 6.5.8 Relational operators
  - [ ] 6.5.9 Equality operators
  - [ ] 6.5.10 Bitwise AND operator
  - [ ] 6.5.11 Bitwise exclusive OR operator
  - [ ] 6.5.12 Bitwise inclusive OR operator
  - [ ] 6.5.13 Logical AND operator
  - [ ] 6.5.14 Logical OR operator
  - [ ] 6.5.15 Conditional operator
  - [x] 6.5.16 Assignment operators
  - [ ] 6.5.17 Comma operator

PR #297, opened 2022-10-04
Add initial support for conditional expressions

reima:
Conditional expressions can be parsed and code generation works, but only for arithmetic types and the void type.

Relates to #207 and #208.

These cases from 6.5.15 are not yet supported (because I don't know how to properly handle these cases):
* both operands have the same structure or union type
* both operands are pointers to qualified or unqualified versions of compatible types
* one operand is a pointer and the other is a null pointer constant
* one operand is a pointer to an object type and the other is a pointer to a qualified or unqualified version of `void`


kant2002:
Last three bullet points probably you can handle in Lower method. Just test for these special cases and  provide alternative expression as needed. Null is 0 and you can wrap that expression in TypeCastExpression(CTypeSystem.NativeInt) for example

ForNeVeR (approved):
Awesome, thanks!

PR #298, opened 2022-10-04
Fix executed test count for single case run

reima:
I noticed that running a single integration test with `Run-Tests -TestCaseName ...` prints `27 tests have been executed successfully` (or whatever the total test count is at the time).

I changed the PowerShell script to instead print the actual number of successful tests (in this case 1).

While I was at it, I also added support for proper singular forms of the sentences in the failed and success cases.


ForNeVeR (approved):
Thanks!

PR #304, opened 2022-10-10
Add missing approved files

kant2002:
Big question why we have tests passed without these files present?


ForNeVeR:
The real issue was that the test methods were missing the `Task` return type, and so Xunit wasn't even checking their results.

ForNeVeR (approved):
Thanks!

PR #307, opened 2022-10-10
Add rand and srand

kant2002:


Cesium.Runtime/StdLibFunctions.cs
```diff
@@ -5,6 +5,9 @@ namespace Cesium.Runtime;
 /// </summary>
 public unsafe static class StdLibFunctions
 {
+    public const int RAND_MAX = 0x7fff;
```
ForNeVeR, line 8: In GCC, this is `0x7FFFFFFF`. I think that would be better, since it is more in line with `Random.Next()`.

Cesium.Runtime/StdLibFunctions.cs
```diff
@@ -15,4 +18,14 @@ public static void Exit(int exitCode)
...
     }
+
+    public static int Rand()
+    {
+        return shared.Next(RAND_MAX);
+    }
+
+    public static void SRand(uint seed)
+    {
+        shared = new Random(unchecked((int)seed));
```
ForNeVeR, line 29: We compile stdlib in unchecked mode by default (and have no plans to change that), so explicit `unchecked` isn't necessary.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #308, opened 2022-10-10
Add non-standard function

kant2002:
These functions appears in the tutorial, so I would like to have them

### DIFF
--- a/Cesium.Runtime/UniStdFunctions.cs
+++ b/Cesium.Runtime/UniStdFunctions.cs
@@ -4,7 +4,7 @@ public static class UniStdFunctions
 {
     public static int Sleep(uint duration)
     {
-        Thread.Sleep((int)(duration * 1000)F);
+        Thread.Sleep((int)(duration * 1000));
         return 0;
     }
 }


Cesium.Compiler/stdlib/conio.h
```diff
@@ -0,0 +1,8 @@
+__cli_import("Cesium.Runtime.ConioFunctions::KbHit")
+int kbhit(void);
+
+__cli_import("Cesium.Runtime.ConioFunctions::GetCh")
+int getch(void);
+
+__cli_import("Cesium.Runtime.ConioFunctions::Cls")
+void cls(void);
```
ForNeVeR, line 8: I wasn't able to find this function in any `conio.h` implementation I searched. Where is this from?

ForNeVeR (changes requested):
I'm not against including some of the widely-known non-standard functions, but we need to be strict of what we include. For now, `cls` is questionable, everything else is fine.

Cesium.Compiler/stdlib/conio.h
kant2002: That’s my replacement for `system(“cls”)`. Looking backwards I better provide system function with this functionality. Or maybe you know better way to clear console screen?

Cesium.Compiler/stdlib/unistd.h
```diff
@@ -0,0 +1,2 @@
+__cli_import("Cesium.Runtime.UniStdFunctions::Sleep")
+int sleep(double);
```
ForNeVeR, line 2: [Normally](https://www.man7.org/linux/man-pages/man3/sleep.3.html), `sleep` is `unsigned sleep(unsigned seconds)`. Are you sure we want `double` here? Yes, it looks more useful honestly, but in reality it can break compilation of certain programs.

ForNeVeR (changes requested):

Cesium.Compiler/stdlib/unistd.h
kant2002: Hmm. It seems to be I found weird sample. it has `sleep(0.01)` and `void main()`. Let me look what on Windows.

ForNeVeR: Sorry, are you still looking at that?

kant2002: Forget about this part.

ForNeVeR (approved):
Thanks!

PR #311, opened 2022-10-12
Rearrange documentation to be more interesting to new people

kant2002:
Put usage closer to the top
Put Implementation Status lower since we are relatively good for now, and


ForNeVeR:
I've decided to mess with it a bit more, and also drop Sonar (I can't see the value of it currently).

ForNeVeR (approved):
Thanks for help!

PR #313, opened 2022-10-12
Consume result of function call if not needed

### LINKED ISSUE #312: Results should be consumed for function calls
Following is simplest way to reproduce problem.
```
int test() { return 0; }

int main() {
    test(); // produce invalid codegen
    int x = test(); // produce correct codegen
}
```

Basically, we should pop result of the call if we do not use it.

kant2002:
Closes #312


ForNeVeR (approved):
Thanks!

PR #315, opened 2022-10-12
Add missing rebuild target for VS

kant2002:


Cesium.IntegrationTests/Cesium.IntegrationTests.proj
```diff
@@ -7,5 +7,6 @@
     </ItemGroup>
 
     <Target Name="Build" />
+    <Target Name="ReBuild" />
```
ForNeVeR, line 10:
```suggestion
    <Target Name="Rebuild" />
```

(.NET SDK calls it `Rebuild`, in this case; let's be consistent here)

ForNeVeR (approved):
Thanks!

PR #316, opened 2022-10-13
Switch statement

kant2002:
A lot of noise comes from fact that I distiguish between constant literals and constant expressions. We incorrectly name ConstantExpresion the expression which works with constants, but that's more broad concept. In the spec names are `constant` and `constant-expression` and because we have suffix Expression to all expression we have a clash.

Closes https://github.com/ForNeVeR/Cesium/issues/74 and superseeds https://github.com/ForNeVeR/Cesium/pull/139


ForNeVeR (approved):
Thanks!

PR #320, opened 2022-10-13
Add snake sample

kant2002:
This sample need slightly modified version of sleep, to be able play okayish. Proper C way is to have `nanosleep` which require interop of structs between C and Cesium runtime which is not there yet. I would like to temporary have non-standard `sleep` function, because it allow us to have tech demo


ForNeVeR (approved):
Thanks!

The game is pretty hard though :(

PR #324, opened 2022-10-17
Add support for two dimensional arrays

kant2002:
- Array access not medeled as pointers plus offset.


kant2002:
Operator between keyboard and display made mistake!

ForNeVeR (approved):
Thanks!

PR #327, opened 2022-10-19
Allocate global array variables on the heap

kant2002:
In previous PR I allocate global array variables using localalloc in the module constructor. That's obviously does not work, but small integration tests does nto catch that. Only after I start running GOL I discover that issue. For now I allocate on the heap and only in .NET 6 configuration. following issues
- Provide netstandard implementation for memory allocation (basically port NativeMemory)
- Codegen for memory deallocation if module is unloaded.

**After merge (F):**
- [ ] open a new issue about module deinitializers

### DIFF
--- a/Cesium.Runtime/RuntimeHelpers.cs
+++ b/Cesium.Runtime/RuntimeHelpers.cs
@@ -53,4 +53,13 @@ public static unsafe class RuntimeHelpers
     {
         System.Environment.Exit(exitCode);
     }
+
+    public static void* AllocateGlobalField(uint size)
+    {
+#if NETSTANDARD2_0
+        throw new NotImplementedException();
+#else
+        return NativeMemory.Alloc(size);
+#endif
+    }
 }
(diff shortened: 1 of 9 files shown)


Cesium.Runtime/RuntimeHelpers.cs
```diff
@@ -53,4 +53,13 @@ public static void Exit(int exitCode)
     {
         System.Environment.Exit(exitCode);
     }
+
+    public static void* AllocateGlobalField(uint size)
+    {
+#if NETSTANDARD2_0
+        throw new NotImplementedException();
```
ForNeVeR, line 60: Won't `Marshal.AllocHGlobal` work for this purpose?

ForNeVeR (changes requested):

ForNeVeR:
Seeing this module initializer makes me think what to do when we unload the library. I really want our assemblies to be good citizens in the CLI world (and thus work with multiple load/unloads), but AFAIK there's no such thing as "module deinitializer" in CLR. We'll have to invent something. I'll open a new issue after merging this.

ForNeVeR (approved):
Thanks!

PR #329, opened 2022-10-19
Printf return int

kant2002:
Some code depends on this.

### DIFF
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -101,9 +101,9 @@ public unsafe static class StdIoFunctions
             formatStartPosition = formatString.IndexOf('%', currentPosition);
         }
 
-        string reminderString = formatString.Substring(currentPosition);
-        Console.Write(reminderString);
-        return consumedBytes + reminderString.Length;
+        string remainderString = formatString.Substring(currentPosition);
+        Console.Write(remainderString);
+        return consumedBytes + remainderString.Length;
     }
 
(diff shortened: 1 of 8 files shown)


Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -24,12 +24,12 @@ public static void PutS(byte* str)
...
     }
 
-    public static void PrintF(byte* str, void* varargs)
+    public static int PrintF(byte* str, void* varargs)
     {
         var formatString = Unmarshal(str);
         if (formatString == null)
         {
-            return;
+            return 1;
```
ForNeVeR, line 32: According to C17 standard,
> The `printf` function returns the number of characters transmitted, or a negative value if an output or encoding error occurred.
```suggestion
            return -1;
```

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -84,6 +84,7 @@ public static void PrintF(byte* str, void* varargs)
         }
 
         Console.Write(formatString.Substring(currentPosition));
+        return 0;
```
ForNeVeR, line 87: Getting a proper return code is tricky. You can make this `return 1` and open a follow-up issue to make the implementation to return a correct number.

ForNeVeR (changes requested):

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #330, opened 2022-10-19
Allow pointers as global variables

kant2002:


ForNeVeR (changes requested):
Could you please add a test? Thanks!

ForNeVeR (approved):
Thanks!

PR #341, opened 2022-10-23
Support static global variables

### LINKED ISSUE #335: Support internal linkage for global objects
See this code example from #75:
> ```c
> static const unsigned char sha1_padding[64] = …
> ```

What does `static` means here?

After careful reading of section **6.2.2 Linkages of identifiers** of the C17 standard draft, I've came to the following understanding:
1. The file-scoped objects (aka "global variables") may me have either external or internal linkage.
2. `static` file-scoped object may not be accessed outside of its translation unit using its name (so we have to either mangle such objects or hide them in static classes).
3. `extern` file-scoped object _may_ be accessed outside of its translation unit.
4. By default, if no `static` is clarified, the object is accessible by other translation units.

But this all is really tricky, and the difference between a declaration and a definition of a file-scoped object makes me feel really lost.

For now, for the purposes of #75, we'll have to support adding the `static` word to the file-scoped identifiers to allow to effectively hide them from other translation units (i.e. scope them to some static classes or something like this). Two translation units should be able to define a `static` file-scoped variable with the same name, and they shouldn't clash.

ForNeVeR:
Closes #335.


Cesium.Compiler/Compilation.cs
```diff
@@ -83,7 +82,8 @@ private static async Task GenerateCode(AssemblyContext context, string inputFile
         if (lastUnprocessedToken.Kind != CTokenType.End)
             throw new ParseException($"Excessive output after the end of a translation unit {inputFilePath} at {lexer.Position}.");
 
-        context.EmitTranslationUnit(translationUnit);
+        var translationUnitName = Path.GetFileNameWithoutExtension(inputFilePath);
+        context.EmitTranslationUnit(translationUnitName, translationUnit);
```
kant2002, line 86: What if we have two filenames in different folders?

ForNeVeR: Good catch. We'll have to invent something to turn them apart in such case.

ForNeVeR: Covered by #345.

PR #347, opened 2022-10-23
Add support for array of structs

### LINKED ISSUE #232: Complex type array support

kant2002:
Closes #232

## TODO [F]
- [x] finish the implementation of `StructType`'s size calculation
- [x] monitor new TODOs in the code


ForNeVeR (approved):
Thanks!

Cesium.CodeGen/Ir/Types/StructType.cs
```diff
@@ -46,5 +46,5 @@ public TypeDefinition Emit(string name, TranslationUnitContext context)
     public TypeReference Resolve(TranslationUnitContext context) =>
         context.GetTypeReference(this) ?? throw new CompilationException($"Type {this} was not found.");
 
-    public int SizeInBytes => throw new WipException(232, $"Could not calculate size for {this} yet.");
+    public int SizeInBytes => Members.Sum(_ => _.Type.SizeInBytes);
```
ForNeVeR, line 49: But now wait a minute. The actual memory layout of a `struct` includes a lot of various paddings. I don't believe this is correct even for simple cases like `struct Foo { int a; char b; }`

ForNeVeR (changes requested):

ForNeVeR:
We've decided to just add Pack = 1 for our structures, and so this math will start working.

See how the types for constant pool are generated if you want to know how to enable Pack = 1 with Cecil.

ForNeVeR (approved):
Thanks!

ForNeVeR:
Integration tests have failed because we cannot determine proper size of a pointer type now :(

Let's go back to the drawing board, this is too much for me.

ForNeVeR:
Assigning to myself, I'll try to start working on #132 which is required to resolve the question here.

ForNeVeR (approved):
Thank you!

Took some time, but finally I am satisfied by the result.

PR #348, opened 2022-11-08
Variables in for/while-loops and switch

### LINKED ISSUE #205: for loop: support variables
Currently, the `for` loop doesn't support variable declarations in the scope of the loop. It should support them.

Look for the number `205` in the code to find clues to implement this feature.

DavTen:
Closes #205

### DIFF
--- a/Cesium.CodeGen/Contexts/LoopScope.cs
+++ b/Cesium.CodeGen/Contexts/LoopScope.cs
@@ -17,14 +17,34 @@ internal record LoopScope(IEmitScope Parent) : IEmitScope, IDeclarationScope
     public TranslationUnitContext Context => Parent.Context;
     public MethodDefinition Method => Parent.Method;
 
+    private readonly Dictionary<string, IType> _variables = new();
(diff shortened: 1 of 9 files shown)


kant2002:
If you probably notice this look like special cases of other features like nested block definitions. Can you also uncomment https://github.com/ForNeVeR/Cesium/blob/main/Cesium.IntegrationTests/declaration_scopes.ignore.c and check that this and more generic case of couple nested blocks would work. If we decouple declaration scopes from statements we would be able to have all such cases covered

DavTen:
Okay, accepted. I'll pay attention to more generic case. It seems that variables are better handle around CompoundStatement, that would include all other cases. What are you think? @kant2002

kant2002:
@DavTen I think `CompoundStatement` should be fine. You may take into account that switch by itself does not create scope IIRC. So you may need introduce special scopeless CompoundStatement either as separate class, or via additional property

ForNeVeR (approved):
Thanks!

ForNeVeR:
You may proceed with related refactorings of `CompoundStatement` in a new PR if you wish, but this is already good enough to be merged.

PR #356, opened 2022-12-11
CodeGen: implement the wide architecture set

### LINKED ISSUE #354: Wide architecture support
In scope of #132, we want to add the **wide** architecture support, which will use wide pointers (64-bit) everywhere.

This will require some work in the standard library to properly support those (right now, I am thinking of implementing some generic type over pointers with separate casts from normal and from wide pointers, to not recompile the stdlib for every arch separately).

When implementing, make sure to eliminate every instance of substring `#354` in compiler's code and documentation.

ForNeVeR:
Closes #354.

## TODO
- [x] Documentation & planning
  - [x] Allow raw pointers as function parameters?
  - [x] `CPtr` for structure fields
  - [x] `CFuncPtr` for function pointers in structure fields
- [x] Make sure all the cases described in the documentation are covered by the added tests
- [x] Merge with `main`
- [x] Review the standard library changes
  - Most of them may be unnecessary since we no longer change the pointers in signatures
- [x] Fix the tests
- [x] Add tests on interop
  - [x] Call .NET function accepting a raw pointer from wide arch
  - [x] Call .NET function accepting a `(C|F|V)Ptr` from wide arch
- [x] present pointers as ~~`long`~~ `CPtr<T>` in codegen on Wide architecture
  - [x] only in structures; see `Cesium.CodeGen.Tests.ArchitectureDependentTypeTests.StructWithPointer`
  - [x] test for ptr to ptr
- [x] Present function pointers as `FPtr` on Wide architecture
  - also only in structures
- [x] Accommodate #440 here
- [x] Add a special `InteropType`, CLI imports should produce functions with interop types as arguments
  - [x] `InteropType` knows how to convert a normal type to CLI type
- [x] At this point, all the tests should become green
  - [x] Well, apparently they didn't, but I should make them so
- [x] `CodeGenNetInteropTests`: add actual run test for `CPtr` interop
- [x] Add a test for array addressing in the wide arch
- [x] Document the function pointer solution
- [x] standard library enhancement
  - [x] create a `CPtr<T>` type
  - [x] add implicit conversions
  - [x] add codegen support
  - [x] support `CPtr` interop
- [x] `FPtr` interop
- [x] integration test
- [x] Check what's going with `__nint`
  - Perhaps just don't support it in the wide architecture?
- [x] TODO sweep
- [x] Make sure everything discussed in #354 is implemented
- [x] Final review

### Before merge
- [x] TODO monitoring
  - [x] comments
  - [x] `ToDo`
- [x] Create an issue on reverse `*Ptr` interop (e.g. as a function return value, or by pointer)
- [x] Create an issue to turn `nint`/`nuint` into `long`/`ulong` on the Wide arch


Cesium.Runtime/FPtr.cs
```diff
@@ -0,0 +1,14 @@
+namespace Cesium.Runtime;
+
+/// <summary>A class encapsulating a C function pointer.</summary>
+public readonly unsafe struct FPtr<TDelegate> where TDelegate : Delegate // TODO: Think about vararg and empty parameter list encoding.
```
kant2002, line 4: Maybe `FuncPtr`?

Cesium.Runtime/CPtr.cs
```diff
@@ -19,7 +19,7 @@ private CPtr(long value)
 
 /// <summary>A class encapsulating an object pointer.</summary>
 /// <typeparam name="T">Type this pointer may be resolved to.</typeparam>
-public unsafe readonly struct CPtr<T> where T : unmanaged
+public readonly unsafe struct CPtr<T> where T : unmanaged
```
kant2002, line 22: And maybe `VoidPtr`. I do like CPtr, but it sounds too broad for me. But that's just remark. I do have only vague feeling here.

kant2002:
Not sure if somebody think about it, so I leave it.
Function pointers IMO we whould represent as `RuntimeMethodHandle`. When convert to `void*` we can convert to `nuint` or `CPtr` and then back. For calls of this method we can use `calli`.
Approximately like this. 
https://github.com/dotnet/runtime/blob/476bce3456006245bdade1fc38a1a300b10f91e8/src/mono/mono/tests/calli_sig_check.il#L13-L21

ForNeVeR:
My general idea is to preserve the function pointers as function pointers, when possible. And fall back to various strange things when not.

But importantly, not every function may be presented by a `RuntimeMethodHandle`: if you got a native function, it may not have a "method" per se.

Cesium interop with native code is a bit of underspecified right now (and that's a pretty generous statement already), but it's a thing we'll want to get better at. Eventually. So, our current design should at least consider how it will work in the future.

PR #357, opened 2023-04-10
Add tests for pointer operations

kant2002:
That's taken from #356 in attempt to take what's good and unblock other work if possible


Cesium.Compiler/Properties/launchSettings.json
```diff
@@ -0,0 +1,20 @@
+{
+    "profiles": {
+        "Cesium.Compiler": {
+            "commandName": "Project",
+            "commandLineArgs": "C:\\d\\github\\Cesium\\Cesium.IntegrationTests\\stdlib/iso646.c -o C:\\d\\github\\Cesium\\Cesium.IntegrationTests/bin/out_cs.exe"
```
ForNeVeR, line 5: That's a hardcode I'd like us to avoid. I have nothing against adding some kind of test launchsettings file, though. Maybe let's generalize this a bit and maybe point to our samples directory?

ForNeVeR (changes requested):

Cesium.Compiler/Properties/launchSettings.json
kant2002: I have to take a look how to do that properly. For now I just remove that file. anyway I add it by mistake.

ForNeVeR (approved):
Thanks!

PR #358, opened 2023-04-10
Add support for #pragma once

kant2002:


Cesium.Compiler/FileSystemIncludeContext.cs
```diff
@@ -6,28 +6,39 @@ public class FileSystemIncludeContext : IIncludeContext
...
+    }
+
+    public TextReader OpenFileStream(string filePath) => new StreamReader(filePath);
+
+    public bool CanIncludeFile(string filePath)
+    {
+        return !visitedFiles.Contains(filePath);
+    }
+
+    public void RegisterPragmaOnceFile(string filePath)
```
ForNeVeR, line 40: _(Minor)_ I propose to adjust the names here. Let's call the `#pragma once` feature an "internal include guard", and then:
```suggestion
    public bool ShouldIncludeFile(string filePath)
    {
        return !visitedFiles.Contains(filePath);
    }

    public void RegisterGuardedFileInclude(string filePath)
```

Cesium.Compiler/FileSystemIncludeContext.cs
```diff
@@ -6,28 +6,39 @@ public class FileSystemIncludeContext : IIncludeContext
 {
     private readonly string _stdLibDirectory;
     private readonly string _currentDirectory;
+    private readonly List<string> visitedFiles = new();
```
ForNeVeR, line 9: _(Minor)_ I suggest a rename:
```suggestion
    private readonly List<string> _guardedIncludedFiles = new();
```

ForNeVeR (approved):

ForNeVeR (approved):
Thanks!

PR #359, opened 2023-04-11
Accept O3 and Wall

kant2002:
For now this parameters do nothing. If we accept them even if we do nothing, that simplify using Cesium when invoking from Make. For example you can CC=/usr/bin/cesium and let Make do it's magic. Eventually

### DIFF
--- a/Cesium.Compiler/Arguments.cs
+++ b/Cesium.Compiler/Arguments.cs
@@ -49,4 +49,10 @@ public class Arguments
     [Option("runtime", HelpText = "Sets path to Cesium C Runtime assembly")]
     public string? CesiumCRuntime { get; init; }
 
+    [Option('O', HelpText = "Set optimization level to 3")]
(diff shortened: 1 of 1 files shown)


Cesium.Compiler/Arguments.cs
```diff
@@ -49,4 +49,10 @@ public class Arguments
     [Option("runtime", HelpText = "Sets path to Cesium C Runtime assembly")]
     public string? CesiumCRuntime { get; init; }
 
+    [Option('O', HelpText = "Set optimization level to 3")]
```
ForNeVeR, line 52:
```suggestion
    [Option('O', HelpText = "Set the optimization level")]
```

ForNeVeR (approved):

ForNeVeR:
Thanks!

I'll open an issue for optimization. For `O > 1`, we could just enable assembly optimization flag.

Issue #350, opened 2022-11-13
Improve IL verifiability


ForNeVeR:
Currently, ILVerify reports several errors in our programs. Namely, considering `array/multidimensional.c` integration test:
```
$ dotnet tool install --global dotnet-ilverify
$ pwsh Cesium.IntegrationTests/Run-Tests.ps1 -TestCaseName array/multidimensional.c
$ cd Cesium.IntegrationTests/bin
$ ilverify out_cs.exe -r "C:\Program Files\dotnet\shared\Microsoft.NETCore.App\7.0.0\*.dll" -r .\Cesium.Runtime.dll
[IL]: Error [StackUnexpected]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::.cctor()][offset 0x00000006][found Native Int][expected Int32] Unexpected type on the stack.
[IL]: Error [Unverifiable]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::main(int32, uint8**)][offset 0x00000006] Instruction cannot be verified.
[IL]: Error [InitLocals]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::main(int32, uint8**)][offset 0x00000009] initlocals must be set for verifiable methods with one or more local variables.
[IL]: Error [UnmanagedPointer]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::main(int32, uint8**)][offset 0x00000009] Unmanaged pointers are not a verifiable type.
[IL]: Error [StackByRef]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::main(int32, uint8**)][offset 0x00000020][found Native Int] Expected ByRef on the stack.
[IL]: Error [StackUnexpected]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::<SyntheticEntrypoint>(string[])][offset 0x00000002][found Native Int][expected Int32] Unexpected type on the stack.
[IL]: Error [InitLocals]: [W:\Projects\Cesium\Cesium.IntegrationTests\bin\out_cs.exe : .<Module>::<SyntheticEntrypoint>(string[])][offset 0x0000000A] initlocals must be set for verifiable methods with one or more local variables.
[...]

PR #360, opened 2023-04-11
Add support for named structs

### LINKED ISSUE #218: Support named structures
As of now, this isn't supported:
```c
struct foo {
  int x;
};

struct foo xx;
```

It should be.

Look for the number `218` in the code to find clues to implement this feature.

kant2002:
Closes #218


ForNeVeR (changes requested):
The example from #218 doesn't seem to work; this yields a parse exception.
```
struct Foo
{
    int x;
};

int main(void) {
    struct Foo foo;
    return 0;
}
```

ForNeVeR:
Also I should note that your solution doesn't seem to follow the C standard. I would accept such a solution, though (and would open a new issue to improve it in the future). But let's make the local declaration parser standard-compliant at least :)

What's up with it being not standard-compliant?

The standard essentially introduces two "namespaces" for types: one namespace for `typedef`s, and a second one for named type declarations such as `struct Foo` or `union Uoo`.

`typedef`s may be referred by their given name, while named types should be only referred by their category (`struct` or `union`) and a name, such as `struct Foo`.

Also, it's possible to have totally different things named `Foo` and `struct Foo` in your program:
```c
struct Foo { int A; };
typedef struct { int B; } Foo;
int main(void) {
  struct Foo a;
  Foo b;
  a.A = 0;
  b.B = 0;
}
```

Ineterestingly, it's _disallowed_ to declare both `union Foo` and `struct Foo` at the same translation unit. Only `typedef … Foo` and `struct` (or `union`) `Foo`.

I think we can for now ignore this kind of name collisions (and open a future improvement issue for that), but we should _at least_ parse valid C programs that refer to named types as `struct X`.

ForNeVeR:
Is Windows test failure caused by #350?

kant2002:
No. I did fix error mentioned in #350 in #362 but issue is not related to verification failures

ForNeVeR (approved):
Thanks!

PR #362, opened 2023-04-25
Fix failing multidimecional.c test

kant2002:
- varible b has incorrect dimensions leading to out of bounds error
- syntetic entry point do not cleaunp args
- and fix string allocation, if we have non-English parameters we can truncate args

### DIFF
--- a/Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
+++ b/Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
@@ -186,7 +186,7 @@ internal class FunctionDefinition : IBlockItem
         syntheticEntrypoint.Body.Variables.Add(exitCode);
 
         var instructions = syntheticEntrypoint.Body.Instructions;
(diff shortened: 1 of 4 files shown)


Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
```diff
@@ -186,7 +186,7 @@ public void EmitTo(IEmitScope scope)
         syntheticEntrypoint.Body.Variables.Add(exitCode);
 
         var instructions = syntheticEntrypoint.Body.Instructions;
-        var atExitLdLocExitCode = Instruction.Create(OpCodes.Ldloc_S, exitCode);
+        var atExitLdLocExitCode = Instruction.Create(OpCodes.Nop);
```
ForNeVeR, line 189: Let's get rid of this variable and `ldloc` altogether? I think it's a remnant of an earlier implementation that was entering the `try` block in this function.

ForNeVeR (changes requested):
I'll try applying my changes and see what happens.

ForNeVeR (approved):
Thanks!

PR #363, opened 2023-04-26
Fix puts signature

### LINKED ISSUE #156: Discard of function return values leads to invalid CLI programs
It turns out that Cesium can compile this simple program, but it won't run:
```c
int foo() { return 1; }
int main() {
  foo();
  return 0;
}
```

Seemingly, it gets confused by an unused return value of the function.

Exterminate `// TODO[#156]` when solving this case. It is a problem of integration tests and e.g. a `puts` function.

kant2002:
Previously was implemented as part of #312

Closes #156


ForNeVeR (approved):
Ah, great. Thanks!

PR #364, opened 2023-04-26
Add typedef resolution in function parameters

kant2002:
This allow implement FILE as opaque handle
Internally it would be index on the internal structure to streams.

### DIFF
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -14,8 +14,8 @@ public unsafe static class StdIoFunctions
     record class StreamHandle
     {
         public required string FileMode { get; set; }
-        public TextReader? Reader { get; set; }
-        public TextWriter? Writer { get; set; }
+        public Func<TextReader>? Reader { get; set; }
+        public Func<TextWriter>? Writer { get; set; }
     }
 
     private static List<StreamHandle> handles = new();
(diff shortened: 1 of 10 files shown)


Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -23,6 +52,39 @@ public static void PutS(byte* str)
             // return EOF; // TODO[#156]: Uncomment
         }
     }
+    public static int PutChar(byte character)
+    {
+        try
+        {
+            Console.Write((char)character);
+            return 0;
```
ForNeVeR, line 60: According to the standard:
> The `putchar` function returns the character written. If a write error occurs, the error indicator for
> the stream is set and `putchar` returns `EOF`.

```suggestion
            return character;
```

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -23,6 +52,39 @@ public static void PutS(byte* str)
...
+        try
+        {
+            var streamDescriptor = GetStreamHandle(stream);
+            if (streamDescriptor == null)
+            {
+                return -1;
+            }
+
+            streamDescriptor.Writer!.Write((char)character);
+            return 0;
```
ForNeVeR, line 80: Should be same as `putchar`, basically.

```suggestion
            return character;
```

ForNeVeR (changes requested):
Thanks, good work! A couple of small suggestions, though.

ForNeVeR:
Also, could you please resolve the conflict here?

ForNeVeR (approved):
Thanks!

PR #366, opened 2023-04-28
Basic compound initialization

### LINKED ISSUE #75: Compiler: global variables

kant2002:
That's very primitive support for aray initialization, just enough to try first milestone Closes #75

### DIFF
--- a/Cesium.CodeGen/Contexts/AssemblyContext.cs
+++ b/Cesium.CodeGen/Contexts/AssemblyContext.cs
@@ -1,6 +1,7 @@
+using System.Collections;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using System.Text;
+using Cesium.Ast;
 using Cesium.CodeGen.Contexts.Meta;
 using Cesium.CodeGen.Extensions;
 using Cesium.CodeGen.Ir.Types;
@@ -41,7 +42,7 @@ public class AssemblyContext
         return assemblyContext;
     }
 
-    public void EmitTranslationUnit(string name, Ast.TranslationUnit translationUnit)
+    public void EmitTranslationUnit(string name, TranslationUnit translationUnit)
     {
         var nodes = translationUnit.ToIntermediate();
         var context = new TranslationUnitContext(this, name);
@@ -67,7 +68,7 @@ public class AssemblyContext
     public const string ConstantPoolTypeName = "<ConstantPool>";
 
     private readonly Dictionary<int, TypeReference> _stubTypesPerSize = new();
-    private readonly Dictionary<ByteArrayWrapper, FieldReference> _dataConstantHolders = new(new ByteArrayEqualityComparer());
+    private readonly Dictionary<ByteArrayWrapper, FieldReference> _dataConstantHolders = new();
 
     private readonly Lazy<TypeDefinition> _constantPool;
     private MethodDefinition? _globalInitializer;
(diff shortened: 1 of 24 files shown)


Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -168,6 +168,20 @@ public FieldReference GetConstantPoolReference(string stringConstant)
         return field;
     }
 
+    public FieldReference GetConstantPoolReference(byte[] dataConstant)
+    {
+        var hash = dataConstant.GetHashCode().ToString();
```
ForNeVeR, line 173: It's not a good idea to take a hash code of a runtime array. At least this won't get you a reproducible build, and at the worst, it will outright break.

Probably you wanted to get a hash code from the array's contents. And also, mind that the runtime's hash code implementation is never meant to be stable, and may change its results between runs, which will, again, break the reproducibility of builds.

I've written a suggested implementation using SHA256 here when I recognized that we should actually generify the `_stringConstantHolders` to use a byte array as the key (using an appropriate comparer, of course). And then, rename all the related stuff accordingly ("StringConstant" → "DataConstant").

ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -165,12 +165,38 @@ public FieldReference GetConstantPoolReference(string stringConstant)
...
+    {
+        if (dataConstant.Length >= 10)
+        {
+            // Do not attempt to put large arrays in cache.
+            // Most likely they are single use in the application.
+            // No chance that they would be cached by values.
+            var bufferSize = dataConstant.Length;
+            var type = GetStubType(bufferSize);
+            return GenerateFieldForDataConstant(type, dataConstant);
+        }
```
ForNeVeR, line 184: To be honest, I can't see what do we lose if we avoid putting these into the `_dataConstantHolders`. Barely anything, really. We may spend a little more time on compilation (though if we hash the values in `O(N)` and cache the resulting hash values, then there's zero lose, algorithmically), while we may save little by reusing the same data, if it happens.

I'd say let's avoid this optimization for now.

Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -165,12 +165,38 @@ public FieldReference GetConstantPoolReference(string stringConstant)
...
+        else
+        {
+            var hash = string.Join("", dataConstant.Select(_ => _.ToString("X2")));
```
ForNeVeR, line 187: Come on, let's write a proper `IEqualityComparer` for byte arrays instead of this 😸 

That means the type of `_dataConstantHolders` will have to change, since I prefer to see either a `byte[]` or a byte array wrapper (with cached hash code) as the key.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #368, opened 2023-04-29
Add support for hex and octal literals

kant2002:


Cesium.CodeGen/Ir/Expressions/Constants/IntegerConstant.cs
```diff
@@ -44,4 +44,42 @@ public void EmitTo(IEmitScope scope)
...
+                }
+            }
+
+            return true;
+        }
+
+        if (int.TryParse(text, out value))
+        {
+            return true;
+        }
```
ForNeVeR, line 81: I'll change this to pass the invariant culture to all the `TryParse` calls here.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #369, opened 2023-04-29
Improve preprocessor

kant2002:
- Add support for ||, && in preprocessor expressions
- Preprocessor expressions can be wrapper in parethesis ()
- Added #undef


Cesium.Preprocessor/BinaryExpression.cs
```diff
@@ -26,6 +26,10 @@ public BinaryExpression(IPreprocessorExpression first, CPreprocessorOperator @op
                 return firstValue == secondValue ? "1" : null;
             case CPreprocessorOperator.NotEquals:
                 return firstValue != secondValue ? "1" : null;
+            case CPreprocessorOperator.LogicalAnd:
+                return (!string.IsNullOrEmpty(firstValue) && !string.IsNullOrEmpty(secondValue)) ? "1" : null;
```
ForNeVeR, line 30: Normally in C preprocessor, `true` should be "nonzero". In this code, you consider any non-empty values as "true", which is incorrect. Let's add a check for zero?

Also, while we are at it, all the expressions should return `"0"` instead of `null`.

ForNeVeR (changes requested):

ForNeVeR (approved):
Great, thanks!

PR #370, opened 2023-04-29
Add ability to define preprocesor directive via CLI

kant2002:


ForNeVeR (approved):
Thanks!

That was a new information for me that the preprocessor defined are defined as `1` by default. But I've checked that with GCC, and your implementation is correct.

PR #373, opened 2023-04-30
Implement pointer subtraction

### LINKED ISSUE #260: Implement pointer subtraction
Make this test pass `pointers_substraction.ignore.c`
```
int main(int argc, char *argv[])
{
    int a[10];
    int* b = &a[1];
    int* c = &a[8];
    return 6 * (c - b);
}

```

zawodskoj:
Fixes #260 

pointers/pointer_subtraction.c is passing now


Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
```diff
@@ -93,41 +131,16 @@ public override IType GetExpressionType(IDeclarationScope scope)
...
-            {
-                return leftType;
-            }
+            case (PointerType, not PointerType): return leftType;
+            case (not PointerType, PointerType): return rightType;
+            case (PointerType left, PointerType right):
+                Debug.Assert(left.Base.GetSizeInBytes(scope.ArchitectureSet) == right.Base.GetSizeInBytes(scope.ArchitectureSet));
 
-            return rightType;
+                return scope.CTypeSystem.Int; // ptrdiff_t, must be signed, int is "fine" (nint is not numeric anyway)
```
kant2002, line 141: Can we have NativeInt here? Because on 64-bit platforms that would be 64-bit value, but Int is 32bit.

zawodskoj: This is the same as in #374 - nativeint is not supported in GetCommonNumericType, which is used in some places, so there's more to fix before we can use nativeint here

ForNeVeR: Should be supported already (or if not, then should be easy to implement), so let's do it.

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
```diff
@@ -71,6 +51,64 @@ public override IExpression Lower(IDeclarationScope scope)
...
+                var baseSizeLeft = leftPointerType.Base.GetSizeInBytes(scope.ArchitectureSet);
+                var baseSizeRight = rightPointerType.Base.GetSizeInBytes(scope.ArchitectureSet);
+
+                if (baseSizeLeft == null || baseSizeRight == null || baseSizeLeft != baseSizeRight)
+                    throw new CompilationException("Invalid pointer subtraction - pointers are referencing different base types");
+
+                return new ArithmeticBinaryOperatorExpression(
+                    new ArithmeticBinaryOperatorExpression(left, Operator, right),
+                    BinaryOperator.Divide,
+                    new ConstantLiteralExpression(new IntegerConstant(baseSizeLeft.Value))
```
ForNeVeR, line 74: Better use `GetSizeInBytesExpression`: it will also manage to use `sizeof` instruction when necessary (i.e. in dynamic arch when the type size is arch-dependent, e.g. a pointer).

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
```diff
@@ -71,6 +51,64 @@ public override IExpression Lower(IDeclarationScope scope)
...
+            {
+                if (Operator != BinaryOperator.Subtract)
+                {
+                    throw new CompilationException($"Operator {Operator} is not supported for pointer/pointer operands");
+                }
+
+                var baseSizeLeft = leftPointerType.Base.GetSizeInBytes(scope.ArchitectureSet);
+                var baseSizeRight = rightPointerType.Base.GetSizeInBytes(scope.ArchitectureSet);
+
+                if (baseSizeLeft == null || baseSizeRight == null || baseSizeLeft != baseSizeRight)
```
ForNeVeR, line 68: I suggest directly comparing the types instead of comparing their sizes: the current implementation will allow subtracting pointers to different types of identical size, which shouldn't be possible.

ForNeVeR (changes requested):
I believe that this is a step in the right direction.

Aside from mentioned above:
- Could you please add a test into `CodeGenMethodTests`?
- Could you also take a look at the test failures?

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
zawodskoj: Done

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
zawodskoj: Done

Cesium.CodeGen/Ir/Expressions/BinaryOperators/ArithmeticBinaryOperatorExpression.cs
zawodskoj: Rebased and fixed

zawodskoj:
I've added few more tests

ForNeVeR (approved):
Great, thanks!

PR #371, opened 2023-04-29
Add strcpy, strlen, strncpy, strcat, strncat

kant2002:
Depends on #364

PR #374, opened 2023-04-30
WIP: Implement offsetof as builtin

### LINKED ISSUE #276: Offsetof implementation

zawodskoj:
TypeSystemEx prevents this to be nativeint (as it should be), as nativeint is not "numeric" (which should be fixed I think)

Fixes #276

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
+++ b/Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
@@ -36,6 +36,35 @@ internal class FunctionCallExpression : IExpression
 
     public IExpression Lower(IDeclarationScope scope)
     {
+        if (_function.Identifier == "__builtin_offsetof")
+        {
+            if (_arguments.Count != 2 ||
+                _arguments[0] is not IdentifierExpression typeExpr ||
+                _arguments[1] is not IdentifierExpression fieldExpr)
+            {
+                throw new CompilationException($"offsetof: invalid arguments");
+            }
+
+
+            var type = typeExpr.Identifier;
+            var field = fieldExpr.Identifier;
+
+            var resolvedType = scope.ResolveType(new NamedType(type));
+            if (resolvedType is not StructType resolvedStruct)
+            {
+                throw new CompilationException($"Type \"{type}\" is not a struct type.");
+            }
+
+            var fieldType = resolvedStruct.Members.SingleOrDefault(m => m.Identifier == field);
+
+            if (fieldType == null)
+            {
+                throw new CompilationException($"Member \"{field}\" was not found in struct \"{type}\".");
+            }
+
+            return new OffsetOfExpression(resolvedStruct, fieldType);
+        }
+
         var functionName = _function.Identifier;
         var callee = scope.GetFunctionInfo(functionName);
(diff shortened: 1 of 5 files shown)


Cesium.Compiler/stdlib/stddef.h
```diff
@@ -1,8 +1,9 @@
-typedef __nint ptrdiff_t;
+// typedef __nint ptrdiff_t;
```
kant2002, line 1: You would not need this after #371

Cesium.CodeGen/Ir/Expressions/OffsetOfExpression.cs
```diff
@@ -0,0 +1,49 @@
...
+            var = existingVar;
+        }
+        else
+        {
+            var = new VariableDefinition(typeRef);
+            scope.Method.Body.Variables.Add(var);
+        }
+
+        scope.AddInstruction(OpCodes.Ldloca, var);
+        scope.AddInstruction(OpCodes.Ldflda, fieldRef);
```
kant2002, line 41: I think this should be intrinsic which just calculate value of offset, no rely on runtime for that. but.... On the second thought, what if we have .NET type from external library which would like to use. Personally I would still have constant evaluation first, and when we implement interop with .NET we special case these types.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
```diff
@@ -36,6 +36,35 @@ public FunctionCallExpression(Ast.FunctionCallExpression expression)
 
     public IExpression Lower(IDeclarationScope scope)
     {
+        if (_function.Identifier == "__builtin_offsetof")
+        {
+            if (_arguments.Count != 2 ||
```
kant2002, line 41: I think following cases not yet handled 
```
struct foo
{
    int bar;
    int baz;
    int qux;
}

offsetof(struct S, baz);
```

Maybe that require special tricks on the parser, create special construct for passing type name in the functions. Maybe easier target for learning would be implementing `sizeof()` in the parser first, because we already have https://github.com/ForNeVeR/Cesium/blob/6e6136c335f29a03afdd6354e4c0a7932e405029/Cesium.CodeGen/Ir/Expressions/SizeOfExpression.cs

Cesium.CodeGen/Ir/Expressions/OffsetOfExpression.cs
zawodskoj: I think there's no need to dig into implementation-defined details of the runtime, it's better to use what runtime already provides

kant2002: Cesium has their own rules. All struct firstly defined by Cesium and layout is controlled by Cesium

zawodskoj:
Replaced by #379

PR #375, opened 2023-04-30
WIP stupid control flow implementation

zawodskoj:
attempt to make stupid control flow checker

FinalNodes is thing to think about, it is only required in CompoundStatement, but these nodes may depend on continuations inside CompoundStatement and therefore may be not yet resolved at the moment they needed - hence "production-quality" heck called `unsafeNext`


zawodskoj:
I didn't dare running all of the tests, but at least goto/forward.c now passes without extra return

kant2002:
I would say running tests would be required
image

I did not dig deeper, but personally I really would like to have BB (Basic blocks) for CFG analysis. I understand that this is somewhat complicated to implement currently. Other stuff which concerns me is that you introduce concept of Parent for this simple stuff only. I think exists easier way to solve that.

PR #376, opened 2023-05-01
Allow nativeints to be coerced

zawodskoj:
This allows nativeints to participate in few math operations without narrowing to explicit ints/longs/whatever.

#373 and #374 should rebase on this if merged


ForNeVeR:
I consider `NativeUInt` and `NativeInt` as intrinsic types corresponding to C's `size_t` and `ptrdiff_t`, respectively. So, in Cesium, they actually _should_ be the normal arithmetic types, and should participate in arithmetics. (And they are called `__nuint` and `__nint` in C extended syntax for now, which wasn't yet fully implemented and documented at the moment you've opened the PR, so you aren't the one to blame.)

In contrast with, say, `void*` that shouldn't.

Part of the change that is supposed to bring them to arithmetics was done in 1fe86f344521f0a99f05c6c2a947807b1bf4d68c, and the remaining work (choose the right conversion rank) is tracked in #381.

I'm sorry, but I have to reject this change.

PR #377, opened 2023-05-01
Add support for type cast expressions enclosed with parens

zawodskoj:
Expressions like `(size_t) (1 + 2)` are now supported
Type casts to typedef aliases are now supported as well


Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
```diff
@@ -32,9 +33,46 @@ public FunctionCallExpression(Ast.FunctionCallExpression expression)
         _arguments = (IList<IExpression>?)arguments?.Select(e => e.ToIntermediate()).ToList()
                      ?? Array.Empty<IExpression>();
         _callee = null;
+        _castLike = castLike;
     }
 
     public IExpression Lower(IDeclarationScope scope)
+    {
+        if (_castLike)
```
kant2002, line 41: Is is possible that `castLike` is not needed after #365 ? The only concern is `size_t c = (foo)(1, 2);` but `(foo)` is function address expression, so I do not see that it need special handling like `castLike`

kant2002: I look at at closer and understand the problem. I think better way, or maybe that y proference is to introduce new node 
```
record class FunctionCallOrCastExpression(function, arguments);
```

which in the lower phase will try to perform all checks and detect is it function call or type and create new nodes `FuncntionCallExpression` respectively and `CastExpression` respectively and lower them inside. That way IMO we have clearly function calls and clearly cast expression and some muddy syntax details would be encapsulated in one place.

Cesium.Parser/CParser.cs
```diff
@@ -90,7 +90,13 @@ public partial class CParser
         Expression function,
         IToken _,
         ArgumentExpressionList? arguments,
-        IToken __) => new FunctionCallExpression(function, arguments);
+        IToken __)
+    {
+        var castLike = function is ParenExpression { Contents: ConstantLiteralExpression } &&
```
kant2002, line 95: I think this should not be part of the parser, you can have that check in the lower phase too. so condition to have cast is to 
- have identifier in the paren (castLike)
- have identifier to be resolved to type
so if you have 2 checks in one place, this parameter would be not needed.

Cesium.IntegrationTests/type_casts.c
```diff
@@ -0,0 +1,17 @@
...
+    return a + b;
+}
+
+int main(void)
+{
+    int a = (int) 1.0;
+    size_t b = foo(1, 2); // ((size_t)) a;
+    size_t c = (foo)(1, 2); // ((size_t)) a;
+    size_t d = (size_t) a;
+    size_t e = (size_t) (a + b);
```
kant2002, line 14: Can you please add more expressions like this
```
int f = (int)a;
int g = (int)(a + b);
```
Just to check that primitive types are also works.
Whole expression become more complex, so maybe it would be better to just put parts of it in `if`, something like this https://github.com/ForNeVeR/Cesium/blob/940c1cb23d953a5a03ed7629e8ecfa2d10420b44/Cesium.IntegrationTests/array/multidimensional.c#L5-L10

Cesium.Parser/CParser.cs
zawodskoj: The whole reason of having 'castLike' is to avoid ParenExpressions to appear in other parts of AST (they have no value by itself) - at this step we can check whether we have exactly one paren pair here and throw them away

Note that `((some_type)) (expr)` is no longer a valid cast

Cesium.IntegrationTests/type_casts.c
zawodskoj: I added test cases for (int) cast

Cesium.IntegrationTests/type_casts.c
```diff
@@ -0,0 +1,26 @@
...
+
+    size_t d = (size_t) a;
+    size_t e = (size_t) (a + b);
+    if (d + e != 5) return -2;
+
+    int f = (int) a;
+    int g = (int) (a + b);
+    if (f + g != 5) return -3;
+
+    return 42;
```
kant2002, line 25: Can you add the `(struct foo)` and `(void*)x` tests and probably `void (*name_func)(int x)`. If they would be broken due to parsing, you can comment out them for now. I just thinking if `TryGetType` would be enough, or we may need something like https://github.com/ForNeVeR/Cesium/pull/382

zawodskoj: Done

zawodskoj:
Added a lot of tests
Function pointer casts are not working now (lack of correct parsing?)
Added special case for "casting" to void (discarding value)

Cesium.IntegrationTests/type_casts.c
zawodskoj: Using #382 is fine - I can rebase on that when it gets merged

ForNeVeR (approved):
Thanks!

PR #378, opened 2023-05-01
Make argv[0] closer to other platforms

kant2002:


kant2002:
Thanks @kekekeks for idea

ForNeVeR (approved):
For the record, I've done some experiments to figure out if this will work worse than the old way, in cases when the command line starts with `dotnet`.

And it doesn't get worse: both `dotnet run` and `dotnet exec` somehow manage to deliver the assembly name (a `.dll`) as `Environment.GetCommandLineArgs().ElementAtOrDefault(0)`.

So, I agree this is an improvement for various interesting cases such as native CLR hosting.

Issue #355, opened 2022-12-04
Struct layout calculation


ForNeVeR:
To statically determine struct layout (for 32b, 64b and wide architecture sets), we'll need to calculate struct layout.

See `#xxx` number in the code when implementing this.

I am planning to make structures have `Pack = 4 / 8` and `LayoutKind.Explicit` for these architectures.

See [the documentation on pack](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.structlayoutattribute.pack?view=net-7.0).

starvader13:
Hey, This looks interesting, and I want to contribute to this. Can you give me some more details regarding this issue so that I can start working on it asap?

ForNeVeR:
Sure thing. This is a good task for new contributors, because it doesn't require deep knowledge of the existing code.

Here's a more detailed explanation.

We support several CPU architectures (or "architecture sets" as we call them right now): 64-bit, 32-bit, and dynamic. (Let's not talk about **wide** right now since it is not yet implemented, and **dynamic** is out of scope of this task anyway.)

Each bitness requires separate layout of nearly every C structure, because on different architectures, the pointer size are different.

.NET documentation [outlines some basic layout rules](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.structlayoutattribute.pack?view=net-7.0), but I'm not sure if this is the full rule set, and even if then, the wording is very suboptimal, so if you have some domain knowledge in C structure layout, that would be ideal (if not, then you could still try to read the documentation and test on some examples, that would be fine as well).

I want Cesium to support different struct layouts for different architectures, which mean:
1. Set `Pack = 4` or `Pack = 8` in `StructType.cs` based on the target architecture set (already done).
2. Set an equivalent of `[StructLayout(LayoutKind.Explicit)]` for all Cesium-generated structure types (see [`StructType.cs:33`](https://github.com/ForNeVeR/Cesium/blob/2e82febc736a06c0977856127c4df5ddd88e7b66/Cesium.CodeGen/Ir/Types/StructType.cs#L33))[^1].
3. For every struct field, calculate the field layout and generate an equivalent of `[FieldOffset(x)]` in the byte code for said field[^1].
[...]

PR #379, opened 2023-05-01
Implement offsetof as builtin

### LINKED ISSUE #276: Offsetof implementation

zawodskoj:
Closes #276.

Replacement for #374 - I decided to make it different PR as the implementation differs largely.

This implementation introduces builtin to obtain ptr to dirty instance of arbitrary struct.
This allows to write macro appropriate to offsetof features that previous implementation was unable to cover.
For example, `offsetof(foo, bar.baz)` now works.

This PR also uncovers discrepancy between default .net struct layout and msvc struct layout - I was forced to use Sequential struct layout for tests to match msvc offsets (related to #355?)

Do not merge: this PR depends on #377 and should be rebased before
Also there's a lot of testing code and dirty places, which requires some tidying.

## TODO[F]
- [x] Fix the review issues
### After merge
- [ ] Open an issue about using preprocessor in codegen tests
- [ ] Open an issue about the single place for builtins

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenPointersTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenPointersTests.cs
@@ -34,4 +34,110 @@ public class CodeGenPointersTests : CodeGenTestBase
     public void CannotAddPointerTypes() => DoesNotCompile(
         "void foo (int *x) { x = x +x; }",
         "Operator Add is not supported for pointer/pointer operands");
+
+    [Fact]
+    public Task CanUseBuiltinOffsetOfOnDeclaredStruct() => DoTest(
+        "typedef struct { int x; } a; int m() { return &__builtin_offsetof_instance((a*) 0).x; }"
+    );
+
+    [Fact]
+    public Task CanUseBuiltinOffsetOfOnTaggedStruct() => DoTest(
+        "struct a { int x; }; int m() { return &__builtin_offsetof_instance((struct a*) 0).x; }"
+    );
+
+    [Fact]
+    public Task CanUseBuiltinOffsetOfOnBothDeclaredAndStruct() => DoTest(
+        "typedef struct a_s { int x; } a_t; int m() { return (int)&__builtin_offsetof_instance((a_t*)0).x + (int)&__builtin_offsetof_instance((struct a_s*) 0).x; }"
+    );
+
+    [Fact]
+    public Task CanUseBuiltinOffsetOfOnDeepMembers() => DoTest(
+        "typedef struct { int x; } a; typedef struct { a a; } b; int m() { return &__builtin_offsetof_instance((b*)0).a.x; }"
+    );
+
+    [Fact]
+    public void CannotUseBuiltinOffsetOfOnInvalidMembers() => DoesNotCompile(
+        "typedef struct { int x; } a; int m() { return &__builtin_offsetof_instance((a*)0).y; }",
+        "\"a\" has no member named \"y\""
+    );
+
+    [Fact]
+    public void CannotUseBuiltinOffsetOfOnPlainType() => DoesNotCompile(
+        "int m() { return &__builtin_offsetof_instance((int*)0).x; }",
+        "__builtin_offsetof_instance: type \"PrimitiveType { Kind = Int }\" is not a struct type."
+    );
+
+    [Fact]
+    public void CannotUseBuiltinOffsetOfOnPointerType() => DoesNotCompile(
+        "typedef struct { int x; } a; int m() { return &__builtin_offsetof_instance((a**)0).x; }",
+        "__builtin_offsetof_instance: type \"PointerType { Base = Cesium.CodeGen.Ir.Types.StructType }\" is not a struct type."
+    );
+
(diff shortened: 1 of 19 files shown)


ForNeVeR:
> Do not merge: this PR depends on https://github.com/ForNeVeR/Cesium/pull/377 and should be rebased before

In such cases, usually it's better to convert a PR to a draft, for ease of tracking.
image

zawodskoj:
Hah, in Gitlab there's automatic marking of WIP MRs

ср, 3 мая 2023 г., 23:35 Friedrich von Never ***@***.***>:

kant2002:
If you make changes to codegen and that results in broken unit tests, you should run `.\scripts\approve-all.ps1` so you can update baseline. Also that way you can validate that your changes is what you expect.

zawodskoj:
I've rebased this branch and "fixed" all of the tests. There's now an identical layout for all of architectures - forced sequential with 4-byte packing. This change allows to pass a single test which does not actually need to pass anywhere - relying on specific field offsets is UB. I don't know how to test this feature properly, unless we're support specific alignment attributes like `__attribute__ ((aligned (4)))`

I can replace offsetof test with a reasonable check that the pointer is inside adequate bounds, so this PR will not need messing around default pointer alignment.

zawodskoj:
I'm unable to write tests in Cesium.CodeGen.Tests due to lack of preprocessor support in tests

ForNeVeR:
Why don't you use `__builtin_offsetof_instance` in those tests? It seems appropriate.

zawodskoj:
It should work this way, but I think tests should support preprocessor anyway

ForNeVeR:
I believe that the codegen tests should mostly test codegen, and could ignore preprocessor.

There are also integration tests that test everything in complex; I also agree that it would be beneficial to add such tests for this feature as well. Take a look at how `Cesium.IntegrationTests` are structured.

zawodskoj:
I've added some codegen tests
There's already a test on offsetof in IntegrationTests

zawodskoj:
Rebased - commits are cleansed from testing crap

Cesium.CodeGen.Tests/CodeGenPointersTests.cs
```diff
@@ -34,4 +34,110 @@ private static Task DoTest(string source)
...
+        "Cannot resolve type undeclared"
+    );
+
+    [Fact]
+    public void CannotUseBuiltinOffsetOfOnUndeclaredTag() => DoesNotCompile(
+        "int m() { return &__builtin_offsetof_instance((struct undeclared*) 0).x; }",
+        "__builtin_offsetof_instance: struct type \"undeclared\" has no members - is it declared?"
+    );
+
+    // Tests below rely on preprocessor, which is not supported in tests now
```
ForNeVeR, line 88: Note for self: let's open an issue and figure out what to do with these tests.

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
```diff
@@ -36,6 +36,28 @@ public FunctionCallExpression(Ast.FunctionCallExpression expression)
 
     public IExpression Lower(IDeclarationScope scope)
     {
+        if (_function.Identifier == "__builtin_offsetof_instance")
```
ForNeVeR, line 39: Another note for self: let's open an issue on extracting the builtin list to some special place.

Cesium.CodeGen/Ir/Expressions/InstanceForOffsetOfExpression.cs
```diff
@@ -0,0 +1,49 @@
+using Cesium.CodeGen.Contexts;
+using Cesium.CodeGen.Extensions;
+using Cesium.CodeGen.Ir.Expressions.Values;
+using Cesium.CodeGen.Ir.Types;
+using Mono.Cecil.Cil;
+
+namespace Cesium.CodeGen.Ir.Expressions;
+
+internal class InstanceForOffsetOfExpression : IExpression, IValueExpression, IAddressableValue
```
ForNeVeR, line 9:
```suggestion
/// <summary>Takes an address of an instance of the passed type as part of the offsetof expression.</summary>
internal class InstanceForOffsetOfExpression : IExpression, IValueExpression, IAddressableValue
```

Cesium.CodeGen/Ir/Expressions/FunctionCallExpression.cs
ForNeVeR, line 39: Let's document this somewhere.

ForNeVeR (changes requested):
This is great! I've left some notes and will work on them myself in a moment.

ForNeVeR (approved):
Thanks!

PR #382, opened 2023-05-03
Add function for testing that type can be resolved

kant2002:


kant2002:
In case when unresolved type somewhere deep inside the struct we can known what type trigger cascading failures. That’s how original logic works. 

Reporting is still awful but I do not see how can I make proper infrastructure for it.

Cesium.CodeGen/Contexts/TranslationUnitContext.cs
```diff
@@ -75,30 +99,84 @@ internal IType ResolveType(IType type)
...
+                }
+            }
+
+            resolvedType = new StructType(members, structType.Identifier);
+            return isStructResolved;
         }
 
         if (type is FunctionType functionType)
         {
+            bool isFunctionResolved = true;
```
ForNeVeR, line 134: Ok, so there are a lot of warnings on that the variable is unused. Could you remove it then?

ForNeVeR (changes requested):
I apporve this in principle, but could you please add a couple of tests for this new logic?

ForNeVeR:
@kant2002 do you still think we need this? Sadly, I can't remember what the original intent was.

PR #383, opened 2023-05-03
Add support for using global static variables

kant2002:
Case when static variable defined in the function does not handled yet.

### DIFF
--- a/Cesium.CodeGen/Contexts/AssemblyContext.cs
+++ b/Cesium.CodeGen/Contexts/AssemblyContext.cs
@@ -36,6 +36,7 @@ public class AssemblyContext
         var assemblyContext = new AssemblyContext(assembly, module, compilationOptions);
 
         var targetRuntime = compilationOptions.TargetRuntime;
+        //module.AssemblyReferences.Add(targetRuntime.GetSystemAssemblyReference());
         assembly.CustomAttributes.Add(targetRuntime.GetTargetFrameworkAttribute(module));
 
         return assemblyContext;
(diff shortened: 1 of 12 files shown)


ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/AssemblyContext.cs
```diff
@@ -37,6 +37,7 @@ public class AssemblyContext
 
         var targetRuntime = compilationOptions.TargetRuntime;
         assembly.CustomAttributes.Add(targetRuntime.GetTargetFrameworkAttribute(module));
+        module.AssemblyReferences.Add(targetRuntime.GetSystemAssemblyReference());
```
ForNeVeR, line 40: _(Minor)_ This line was removed in #385, but here it's back. I assume we should actually drop that?

The new tests work with it, though, so I assume it doesn't matter, and I merge :)

ForNeVeR (approved):
Thanks!

Cesium.CodeGen/Contexts/AssemblyContext.cs
kant2002: My general policy is to make all changes independent. and this change really was not important for global statics, so I only reduce changeset. It was a slip on my part to put removal in PR.

ForNeVeR: Actually, it seems kinda a rebase/merge artifact rather than an actual change.

PR #385, opened 2023-05-07
Fix compilation for .NET Framework

kant2002:
which was broken for a long time. Because compiler run under TFM net7.0 we canot use `ImportReference(typeof(XXX))` or `ImportReference(typeof(XXX).GetMethod("mymethod"))` since that will bring depedency on S.P.CoreLib to finnal executable, which does not exists in .NET Framework.

Also add another test leg to make sure that it would not regress in the future.

I remove one line in test [/TargetRuntimeTests.cs](https://github.com/ForNeVeR/Cesium/pull/385/files#diff-0b3c15c70dd4a60cb9e5b2a48f0e890eb58fc8dee48810c6f4bc9b20ba043722) because I do not sure what it test since CoreLib is always S.P.CoreLib and after I fix resolving, it always mscorlib due to some Cecil internal stuff.

Also if you run `pwsh -c .\Cesium.IntegrationTests\Run-Tests.ps1 -TestCaseName structs\struct_subscription.c -TargetFramework NetFramework` and open `Cesium.IntegrationTests/bin/out_cs.exe` in ILSpy you would see following

image

This does not prevent anything from running, but indicates that we still work with metadata incorrectly.

### DIFF
--- a/.github/workflows/run-build-and-integration-tests.yml
+++ b/.github/workflows/run-build-and-integration-tests.yml
@@ -18,6 +18,14 @@ jobs:
         environment:
           - ubuntu-latest
           - windows-latest
+        framework:
+          - Net
+        configuration:
+          - Release
+        include:
+          - environment: windows-latest
+            framework: NetFramework
+            configuration: Release
     env:
       DOTNET_NOLOGO: 1
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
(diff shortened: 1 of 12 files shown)


kant2002:
Integration tests did not run for some reason.

kant2002:
Okay. I fix integration tests.

ForNeVeR (approved):
Great, thanks!

PR #386, opened 2023-05-09
Preprocessor parameter replacement eats up whitespaces

kant2002:


ForNeVeR (approved):
I'm not happy with that we have to handle the whitespace this way, but thanks anyway! Maybe we'll consider rolling our own preprocessor (and also parser) implementation, that will handle these more natively, in the future.

PR #387, opened 2023-05-09
Add ability to parse static functions

kant2002:
Most implementation was simply refactoring.

Basically I want make operation closer to how C perceived by developer.
- We declare function if possible. We can declare multiple times. All of that operations happens in Lower phase. This allow combine various properties like storage class until emit happens. CLI import resolved at this stage too.
- Then we define method definition for the function. We simply create new method for all definitions.
- Then emit body (this is not changed)


Cesium.IntegrationTests/function_storage.c
```diff
@@ -0,0 +1,31 @@
...
+    printf("inside storage_declared_worker");
+}
+
+/*
+void storage_defined_worker();
+
+static void storage_defined_worker()
+{
+    printf("inside storage_defined_worker");
+}*/
```
ForNeVeR, line 23: Why is this commented-out? Doesn't it work?

ForNeVeR (changes requested):

Cesium.IntegrationTests/function_storage.c
kant2002: On MSVC it’s fine. In Cesium too.

ForNeVeR: I suspect it breaks the standard then, let's git rid of this particular test?

Perhaps there's no sense in declaring a function with no `static` linkage, and then marking an implementation as `static`.

ForNeVeR (approved):
Thanks!

PR #389, opened 2023-05-10
Remove not needed function which left after rebase

kant2002:
With that function we have two references to System.Runtime when targeting .NET


ForNeVeR (approved):
Thanks!

PR #394, opened 2023-05-13
Add functions for memory management

kant2002:
- aligned_alloc on NetFx is not supported
- Also add system implementation while I was there.

Depends on #393 to pass calloc.c

### DIFF
--- a/Cesium.Runtime/StdLibFunctions.cs
+++ b/Cesium.Runtime/StdLibFunctions.cs
@@ -35,17 +35,27 @@ public unsafe static class StdLibFunctions
     public static int System(byte* command)
     {
         string? shellCommand = StdIoFunctions.Unmarshal(command);
-        var startParameters = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
-            ? new ProcessStartInfo("cmd", $"/c {shellCommand}")
-            : new ProcessStartInfo("/bin/sh", $"-c \"{shellCommand}\"");
-        Process? pa11yCiInstallation = Process.Start(startParameters);
-        if (pa11yCiInstallation is null)
+        if (shellCommand is null)
         {
             return 8 /*ENOEXEC*/;
         }
 
-        pa11yCiInstallation.WaitForExit();
-        return pa11yCiInstallation.ExitCode;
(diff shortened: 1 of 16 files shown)


Cesium.Runtime/StdLibFunctions.cs
```diff
@@ -33,14 +34,68 @@ public static void SRand(uint seed)
 
     public static int System(byte* command)
     {
-        switch (StdIoFunctions.Unmarshal(command))
+        string? shellCommand = StdIoFunctions.Unmarshal(command);
+        var startParameters = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+            ? new ProcessStartInfo("cmd", $"/c {shellCommand}")
+            : new ProcessStartInfo("/bin/sh", $"-c \"{shellCommand}\"");
+        Process? pa11yCiInstallation = Process.Start(startParameters);
+        if (pa11yCiInstallation is null)
```
ForNeVeR, line 42: On Windows at least, you can set a `UseShellExecute` flag instead.

And for the Unix version, let's pass the arguments via `ArgumentList` instead of adding quotes: that's more reliable.

kant2002: ArgumentList is not in netstandard2.0, do we want code split here?

kant2002: and how do you want to use `UseShellExecute` ? Are you want split command line in parts to get executable name ?

kant2002:
Honestly I do not see a way how to make Net FW works without bringing whole Process class implementation in CesiumRuntime

kant2002:
You can ignore last message.

Cesium.Runtime/StdLibFunctions.cs
ForNeVeR: Ok, after reading the docs, I no longer think `UseShellExecute` is of any help for us here.

Also, now I see that we can't use `ProcessStartInfo.ArgumentList` on .NET Standard 2.0. Sigh. Okay, then.

ForNeVeR (approved):
Thanks!

PR #395, opened 2023-05-13
Implement control flow graphs

### LINKED ISSUE #301: Broken program produced.
Cesium produces broken executable from this code. 
```
#include <stdio.h>

// function to check even or not
void checkEvenOrNot(int num)
{
    if (num % 2 == 0)
        // jump to even
        goto even;
    else
        // jump to odd
        goto odd;

even:
    printf("%d is even", num);
    // return if even
    return;
odd:
    printf("%d is odd", num);
    // return; // This statement needed since control flow analysis in Cesium is broken.
}

int main() {
    int num = 26;
    checkEvenOrNot(num);
    return 42;
}
```

If uncomment `return` statement in the `checkEvenOrNot` then magically things start working

zawodskoj:
This PR adds control flow checking via building graph and checking if any statement leading to <terminator> is a ReturnStatement. In case of `void` or `int main` methods - any inappropriate statement is replaced with CompoundStatement containing original statement and synthetic ReturnStatement.

I'll mark this as a draft, but all the tests are passing (including integration tests, even forward.c without redundant return), and the code is fine enough

Closes #301.

### DIFF
--- a/Cesium.CodeGen/Cesium.CodeGen.csproj
+++ b/Cesium.CodeGen/Cesium.CodeGen.csproj
@@ -8,6 +8,8 @@
 
     <ItemGroup>
       <PackageReference Include="Mono.Cecil" Version="0.11.4" />
+      <PackageReference Include="QuikGraph" Version="2.5.0" />
+      <PackageReference Include="QuikGraph.Graphviz" Version="2.5.0" />
     </ItemGroup>
 
     <ItemGroup>
--- a/Cesium.CodeGen/Cesium.CodeGen.csproj
+++ b/Cesium.CodeGen/Cesium.CodeGen.csproj
@@ -9,7 +9,6 @@
     <ItemGroup>
       <PackageReference Include="Mono.Cecil" Version="0.11.4" />
       <PackageReference Include="QuikGraph" Version="2.5.0" />
-      <PackageReference Include="QuikGraph.Graphviz" Version="2.5.0" />
     </ItemGroup>
 
     <ItemGroup>
--- a/Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
+++ b/Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
@@ -297,9 +297,4 @@ internal class FunctionDefinition : IBlockItem
 
         transformed.EmitTo(scope);
     }
-
-    public bool TryUnsafeSubstitute(IBlockItem original, IBlockItem replacement)
-    {
-        throw new NotSupportedException();
-    }
 }
--- a/Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
+++ b/Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
@@ -2,17 +2,13 @@ using Cesium.Ast;
 using Cesium.CodeGen.Contexts;
 using Cesium.CodeGen.Contexts.Meta;
 using Cesium.CodeGen.Extensions;
+using Cesium.CodeGen.Ir.ControlFlow;
 using Cesium.CodeGen.Ir.Declarations;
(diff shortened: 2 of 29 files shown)


Cesium.CodeGen/Ir/BlockItems/FunctionDefinition.cs
```diff
@@ -285,25 +290,351 @@ public void EmitTo(IEmitScope scope)
 
     private void EmitCode(TranslationUnitContext context, FunctionScope scope)
     {
-        var statement = _statement.Lower(scope);
-        statement.EmitTo(scope);
-        if (statement.HasDefiniteReturn == false)
+        var loweredStmt = (CompoundStatement) _statement.Lower(scope);
+
+        var dset = new AdjacencyGraph<CodeBlockVertex, CodeBlockEdge>();
```
kant2002, line 295: Whole process IMO should reside between Lower phase and Emit phase there
https://github.com/ForNeVeR/Cesium/blob/b23fe23a9ca71853dc8590413df476176a3cbcf3/Cesium.CodeGen/Contexts/AssemblyContext.cs#L48-L51

Cesium.CodeGen/Cesium.CodeGen.csproj
```diff
@@ -8,6 +8,8 @@
 
     <ItemGroup>
       <PackageReference Include="Mono.Cecil" Version="0.11.4" />
+      <PackageReference Include="QuikGraph" Version="2.5.0" />
+      <PackageReference Include="QuikGraph.Graphviz" Version="2.5.0" />
```
kant2002, line 12: Not needed depedency

Cesium.CodeGen/Ir/BlockItems/IBlockItem.cs
```diff
@@ -4,8 +4,7 @@ namespace Cesium.CodeGen.Ir.BlockItems;
 
 internal interface IBlockItem
 {
-    bool HasDefiniteReturn => false;
-
+    bool TryUnsafeSubstitute(IBlockItem original, IBlockItem replacement);
```
kant2002, line 7: My current understanding is that this way of tree transformation is somewhat fragile and has cascading effects on codebase where changes in signature of methods propagated across whole codebase even if it's relatively simple process change. I was biten by that even after Lower/EmitTo split and it was not pleasant.

My preference to have some Visitor/Rewriter classes like in Roslyn https://github.com/dotnet/roslyn/blob/303412ce74f70f9ea978bba40705cfcb908d901d/src/Compilers/CSharp/Portable/Binder/TypeofBinder.cs#L47

I think this approach will help separate tree transformation processes. @ForNeVeR not so sure about this IIRC.

kant2002 (commented):
I have feeling which I really cannot back by experience or other knowledge that this is detour from improving architecture. Overall as I told in the chat that this would be solved once Basic Blocks + proper lowering to primitive instructions happens and it would be almost for free.

I know that this can be refactored/reworked as needed, but that invoke deep resistance, since if that was needed it may erase all traces of this contribution. I really don't like that option.

kant2002:
I think this is fix https://github.com/ForNeVeR/Cesium/issues/301

Cesium.CodeGen/Cesium.CodeGen.csproj
ForNeVeR: I am not against it if you use it for some diagnostic purposes, but please make sure to leave at least some instructions on the usage. We can then implement an option to emit a graph, or whatever.

zawodskoj:
Rebased on latest main
All checking code moved to ControlFlowChecker
Handling of statements which should be lowered away is removed
TryUnsafeSubstitute is replaced with a tree-walking inside ControlFlowChecker

ForNeVeR (approved):
Thanks!

I believe this is a step forward for the compiler.

PR #396, opened 2023-05-14
Loop and switch lowering

### LINKED ISSUE #404: Fix issue with `switch (i) <statement>`
See `TODO[#404]` in the integration test code, here:
```c
    /* infinite loop in Cesium */
    /*
    int k = 42;
    switch (i) k = 53;
    if (k != 42) return -2;
    */
```

zawodskoj:
This PR replaces existing emitting code for loops (for, while, do-while) and switch statements.
Loops are now lowered to GenericLoopStatement (which can be lowered further), switch expressions are lowered to if+goto

Closes #404.

## TODO[F]
- [x] Deal with #404

### Before merge
- [x] TODO sweep


zawodskoj:
Two more commits - and now loops are lowered directly to if+goto as switch does

zawodskoj:
Due to 
1. incorrect control flow (hoping for #395) 
2. suboptimal codegen for if-statements
there are lot of dirty CIL that hopefully will be optimized away some day

zawodskoj:
As mentioned in #397, switch statements support bodies without braces (like `switch(1) default: break;`), so CompoundStatement assumption should be reverted.

As mentioned in chat, case statements can be placed in any place inside switch block, not only on "top level" (sample in #399)

Thus we should remove any case processing from SwitchStatement altogether, and provide context (mutable) to case statements, which can be used to declare where the cases really are. I already have prototype for this, but there are problems with scopes and mutable lowering in some places.

We should introduce any sort of extra compilation layer for symbol gluing to move that side effects there. There are few ways to do that, and I think we should stay with the simplest and stupidest way possible now to avoid sinking into endless refactoring (#402)

zawodskoj:
Update: Switch statement now doing it's lowering without attempts to parse sub-statements.

`switch (1) while (1) { default: break }` is now almost possible, but #397 is required to resolve `break`s correctly.

I managed to find hacky way to make it work without touching anything except LabelStatement - there's one more check if LabelStatement is already added a label (to prevent it from trying to add it again).

kant2002:
From what I understand https://github.com/ForNeVeR/Cesium/pull/401 can be handled using new refactoring. I may wait while this lands, or @zawodskoj can just take my bits and put them here.

zawodskoj:
Seems reasonable to move that PR here

zawodskoj:
Rebased #401 on that

ForNeVeR (changes requested):
Thanks!

There's a couple of minor things left, I'll handle them myself.

PR #401, opened 2023-05-15
Add support for most common use of for

kant2002:


zawodskoj:
#396 will supercede this

PR #403, opened 2023-05-16
Fix SetValueExpression emitting

### LINKED ISSUE #409: Fix integration test `switch/no_side_effects.ignore.c`
Right now, it's ignored due to "bad SetValueExpression results in unbalanced stack".

### LINKED ISSUE #411: Handle assign-as-expressions correctly
See `TODO[#411]`, `duff_device.c`:
```c
    // assignments are broken as expressions currently
    // switch(c&3) while((c-=4)>=0) {
    switch(c&3) while(c >= 4) {
      […]
```

zawodskoj:
SetValueExpression has invalid type - type is always equal to type of value to set, but actual emitted code always leave empty stack, so SetValueExpression cannot be used as value and when it is used compiler produces invalid IL.

This PR fixes that - compound assignments now work correctly
This allows `while (i += 1)` and `while (i++)` to be compiled correctly

Fun fact - this code will result in 1 in `i`

```
var i = 0;
i += i += 1;
```

(there's no sequence points, so who cares, but this it worth noting)

- Closes #409.
- Closes #411.

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenOperatorTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenOperatorTests.cs
@@ -36,4 +36,13 @@ public class CodeGenOperatorTests: CodeGenTestBase
 
     [Fact]
     public Task TertiaryOperatorWithoutAssignment() => DoTest(@"int main() { 1 > 2 ? 1 : 2; }");
+
+    [Fact]
+    public Task CompoundAssignment() => DoTest(@"int main() { int i = 0; i += 1; }");
+
+    [Fact]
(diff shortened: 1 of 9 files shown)


zawodskoj:
Sequence-point trickery does not play nicely with GCC (`i = i++` does not actually increment anything)

ForNeVeR:
Please rebase on `main` to proceed with this :)

zawodskoj:
Rebased on main

zawodskoj:
switch/no_side_effects is working now - uncommented

ForNeVeR:
Whoa, this also fixes #411!

ForNeVeR (approved):
Thanks!

Issue #285, opened 2022-09-29
Provide dense packing for the vararg parameters


kant2002:
Currently for each vararg parameter allocated 8 bytes. That's expensive and at the same time disallow passing structs inside functions with varargs. We need allocate exact number of bytes required for varargs buffer.

PR #405, opened 2023-05-18
Implement function pointer calls

zawodskoj:
Varargs are not supported atm - I don't have enough knowledge about them, and they may be target to remake (#285, #196)

## TODO[F]

- [x] TODO sweep


Cesium.CodeGen/Ir/Expressions/IndirectFunctionCallExpression.cs
```diff
@@ -0,0 +1,97 @@
...
+        if (_calleeType.Parameters?.IsVarArg == true)
+        {
+            // TODO: See https://github.com/ForNeVeR/Cesium/issues/285
+            // Using sparse population of the parameters on the stack. 8 bytes should be enough for anybody.
+            // Also we need perform localloc on empty stack, so we will use local variable to save vararg buffer to temporary variable.
+            if (varArgParametersCount == 0)
+            {
+                scope.AddInstruction(OpCodes.Ldnull);
+            }
+            else
```
ForNeVeR, line 49: I don't like this copy-paste from the `FunctionCallExpression`. Let's extract it somewhere, or make them lower into each other, or whatever?

ForNeVeR (changes requested):

zawodskoj:
Rebased

Cesium.CodeGen/Ir/Expressions/IndirectFunctionCallExpression.cs
zawodskoj: I've extracted that into a base class

ForNeVeR (approved):
Thanks!

PR #413, opened 2023-05-24
Extract Lower and EmitTo from IBlockItem

zawodskoj:
This PR moves all codegen-related code from IBlockItem inheritors to a new special classes called BlockItemEmitting and BlockItemLowering

These "visitors" (just a humongous switch-case) will help decouple AST statements from theirs lowered form (a lot of "statements" can't be emitted for example and should be lowered) - now we can introduce the whole new type to designate "lowered" statement (ILoweredBlockItem, for example) without messing with emitting code

Regarding the introduction of the "real visitors" (abstract thingies to override VisitSomething methods) - they do not have much benefits in comparison with just simple switch-case, and they take away flexibility (you should start to think how to return a value from them, or how to make them work with IEnumerable/yield)
They also need a lot more handwritten boilerplate to work, whereas switch-case can be generated with a single hotke.


zawodskoj:
This PR may benefit from #395 in which `HasDefiniteReturn` is removed, effectively making BlockItems simple data objects when both PRs are applied

ForNeVeR:
Could you please rebase after merging of the aforementioned PR #395?

zawodskoj:
Rebased
IBlockItem is now just a useless marker

ForNeVeR (approved):
Thanks!

PR #414, opened 2023-05-30
Unify BinaryOperatorExpression subtypes

zawodskoj:
This PR does few things

- removes BinaryOperatorExpression subtypes and moves all the code to base class
A lot of code (especially emitting code) is either same or can be easily merged
- separates AssignmentExpression from BinaryOperatorExpression and separates assignment operators from BinaryOperator enum
There is no reason to keep these expressions subclassed or linked in any way
- replaces unnecessary lowering stage for a few comparison expressions (>=, <=, !=) with a cleaner emitting code (without any lowering)
- simplifies and future-proofs logical (aka short-circuiting) operators by removing extra `==1`/`==0` opcodes
This also fixes any invalid behavior coming from assumption that `true` is always `1` (which is not)
- adds a few integration tests


ForNeVeR (approved):
Thanks!

PR #415, opened 2023-05-30
Foolproof comma operator

zawodskoj:
This PR fixes CommaExpression emit when left expression returns void - there's nothing to pop from the stack in this case


zawodskoj:
Oops (missed the test which actually spotted incorrect emit of ConstantExpression, i think?), gonna fix that

ForNeVeR (approved):
Thanks!

PR #417, opened 2023-05-30
Remove extra pointer casts

zawodskoj:
This PR removes a huge amount of conv.u/conv.i casts around pointers - they are redundant and it does not matter if they're signed or not. The only available arithmetic operations for pointers are only addition and subtraction - the sign (and any sign extension too) does not matter for them.


zawodskoj:
There's a ton of `ldc 4/ldc 1/mul` pattern around pointer math - we should use constant evaluation whereever possible (#416 will help)

Zero-indexing case (ptr[0]) should be optimized as well

ForNeVeR:
Could you please rebase or merge on top of `main`? This is outdated a bit.

zawodskoj:
Rebased

ForNeVeR (commented):
Thanks!

PR #419, opened 2023-06-09
Static CTypeSystem

zawodskoj:
There's no need to pass that through scopes/contexts, all the types are already fixed

TypeSystemEx can be merged into CTypeSystem, I think


ForNeVeR (approved):
Thanks.

PR #420, opened 2023-06-19
Make subscript operator [] symmetric

### LINKED ISSUE #400: Symmetric []
As soon as [] operator is symmetric, the following code shall be correct:
```C
int direction = 1;
char direction_name = direction["nsew"];
```

Fantoom:
closes #400

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
+++ b/Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
@@ -29,10 +29,24 @@ internal class SubscriptingExpression : IExpression, IValueExpression
     public IExpression Lower(IDeclarationScope scope)
     {
         var expression = LowerExpression(_expression, scope);
-        var elementSize = GetElementSize(expression.GetExpressionType(scope));
+        var index = LowerExpression(_index, scope);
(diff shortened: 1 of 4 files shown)


Cesium.CodeGen/Ir/Expressions/SubscriptingExpression.cs
```diff
@@ -29,10 +29,24 @@ private SubscriptingExpression(IExpression expression, IExpression index)
     public IExpression Lower(IDeclarationScope scope)
     {
         var expression = LowerExpression(_expression, scope);
-        var elementSize = GetElementSize(expression.GetExpressionType(scope));
+        var index = LowerExpression(_index, scope);
+        var expressionType = expression.GetExpressionType(scope);
+        var indexType = index.GetExpressionType(scope);
+
+        switch ((CheckIfTypeIsSubscriptable(expressionType), CheckIfTypeIsSubscriptable(indexType)))
```
kant2002, line 36: Because we are not exhaustive, maybe plain `if` will do the job? or maybe just introduction of variables for `CheckIfTypeIsSubscriptable(...)` calls. I think current logic a bit loaded.

ForNeVeR (approved):
Thanks!

PR #424, opened 2023-06-28
Emit code for enum

kant2002:


Cesium.CodeGen/Ir/BlockItems/EnumeratorDefinition.cs
```diff
@@ -0,0 +1,6 @@
+using Cesium.CodeGen.Ir.Expressions;
+using Cesium.CodeGen.Ir.Types;
+
+namespace Cesium.CodeGen.Ir.BlockItems;
+
+internal record EnumeratorDefinition(string Identifier, IType Type, IExpression Value) : IBlockItem;
```
ForNeVeR, line 6: _(Minor)_ The standard names them "enumeration constants", let's do the same.
```suggestion
internal record EnumConstantDefinition(string Identifier, IType Type, IExpression Value) : IBlockItem;
```

ForNeVeR (approved):
Thanks! The only thing I'd like to change is rename, which I'll do myself.

PR #428, opened 2023-07-04
Add Test Adapter for Cesium Integration tests

kant2002:


ForNeVeR (changes requested):
1. Please include the new projects into the solution file.
2. It looks like the tests aren't runnable, yet? Will you fix that?

ForNeVeR:
Now CI seems broken, will you take a look?

kant2002:
I see following issues with current implementation;
- Locally I have flaky comparison of output from Cesium and VC compiled applications
- CI behave strangely on Ubuntu.

ForNeVeR:
Ok, I'll try to investigate what happens here on the weekend.

PR #442, opened 2023-09-23
Add switch to produce preprocessor result I add -E switch which print preprocessor output to stdout and quit. This request from me to abandon CommandLineParser required infrastructure and handle it myself.

kant2002:


ForNeVeR (approved):
Thanks!

PR #443, opened 2023-10-05
Abstract command line parser

kant2002:
It's hard to make GCC/clang compatible CLI switches with command line parser, this PR make parsing self-contained and I expect that we rewrite it to handwritten parser eventually.


kant2002:
I discover that everything is workign and problem was in me. This is great news. If you think we don't need this infra, then turn this PR off

ForNeVeR (approved):
Thanks!

PR #444, opened 2023-10-06
Improve preprocessor handling

kant2002:
- Expand simple macro definition indefinitely
- Do not trigger #error when inside inactive #if's
- Handle nested #if's
- Handle greater/less then/or_equals
- Make sure that we consume all preprocessor tokens as part of macro evaluation


Cesium.Preprocessor/CPreprocessorOperator.cs
```diff
@@ -6,6 +6,10 @@ internal enum CPreprocessorOperator
 {
     Equals,
     NotEquals,
+    LessOrEqual,
+    GreaterOrEqual,
+    LessThen,
+    GreaterThen,
```
ptasev, line 12: Spelling: `LessThan` GreaterThan`

kant2002: Thank you for keeping me in check

ForNeVeR (approved):
Thanks!

Issue #429, opened 2023-07-15
`...` and `__VA_ARGS__` support


ForNeVeR:

abrahamFerga:
I think this issue can be closed now.

ForNeVeR:
Ah, yeah, closed by #445. Thanks!

PR #445, opened 2023-10-08
#429 Preprocessor: ... and __VA_ARGS__ support

abrahamFerga:
There were a few bugs with the function-like macro, I have tested many cases and this commit seems to fix several things and add support for ... and __VA_ARGS__
Let me know if there is something I am missing here


ForNeVeR (approved):
Thanks!

Issue #126, opened 2022-05-28
Array size check: required in stack array, forbidden in parameter declaration


ForNeVeR:
See `TODO[TBD]` in the code: array size should be required for stack array local declaration (like `char[100] xxx;`), and should be forbidden (or should it? check with the C standard, please) for a parameter declaration (like `void foo(char[100] xxx);`).

Griboedoff:
looks like it you can declare function like this `void foo(char x[100])` and it is equal to `void foo(char x[])`

ForNeVeR:
> looks like it you can declare function like this `void foo(char x[100])` and it is equal to `void foo(char x[])`

Corresponds to my (fading) knowledge of C.

PR #448, opened 2023-10-10
#422 Added support for array initialization without size

### LINKED ISSUE #422: Determine array size based on initializer
As I see, `int arr[] = {1, 2, 3, 4, 5}` is not supported now, but C17 standard does (see §6.7.9). I think we should support it too

abrahamFerga:
I think this should fix #422, I am not sure if I took the right approach, but it is working.
I wanted to add an integration test to make this work, but it seems that sizeof(array) is not working, any ideas on how to test this feature?

One of my changes is close to TODO #126, but I didn't know what to do with it, since it is working in both scenarios mentioned in that issue.

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenArrayTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenArrayTests.cs
@@ -99,4 +99,24 @@ int main() {
     a[1] = 2;
     return a[1];
  }");
+
+    [Fact]
+    public Task GlobalArrayInitializationWithoutSize() => DoTest(@"
+    int ints1[3] = { 1, 2, 3 };
+    int ints2[] = { 1, 2, 1 };
+
+    int main() {
+    return ints1[0] + ints2[2];
+}");
+
+    [Fact]
+    public Task ArrayInitializationWithoutSize() => DoTest(@"int main() {
+    int ints1[3] = { 1, 2, 3 };
+    int ints2[] = { 1, 2, 1 };
+    return ints1[0] + ints2[2];
+}");
+
+    [Fact]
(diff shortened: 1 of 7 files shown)


ForNeVeR:
> I wanted to add an integration test to make this work, but it seems that sizeof(array) is not working

Please open a new issue about that.

> any ideas on how to test this feature?

Add a couple of codegen tests (e.g. to `CodeGenArrayTests.cs`). The tests I would like to see:
```
// global variables
int[3] ints1 = { 1, 2, 3 };
int[] ints2 = { 1, 2, 3 };
int main(void) {
  return ints1[0] + ints2[2];
}
```

```
// local variables
int main(void) {
  int[3] ints1 = { 1, 2, 3 };
  int[] ints2 = { 1, 2, 3 };
}
```

```
// parameter passing
int foo(int[] ints) { return ints[0]; } 
```

These tests will make sure nothing is broken, and the code works in all the expected ways.

> One of my changes is close to TODO #126, but I didn't know what to do with it, since it is working in both scenarios mentioned in that issue.

For now, leave it as-is. That TODO is about the special treatment of cases like `foo (int[300] x)`. We'll deal with it separately.

ForNeVeR (changes requested):
I like the changes, but let's have some new tests, please. Thanks!

abrahamFerga:
Hello there,
I added the tests as you said, thanks for the tips!

ForNeVeR:
JFYI, I'm sorry it takes me a bit too long to review everything these days. I had a small vacation earlier, and am currently getting back on track with the active PRs.

ForNeVeR (approved):
Thanks!

PR #451, opened 2023-10-17
Add isspace

kant2002:


kant2002:
Now I'm curious what locale settings on Github CI is set.

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -129,6 +130,20 @@ public static int FPrintF(void* stream, byte* str, void* varargs)
             streamWriter.Write(formatString.Substring(currentPosition, lengthTillPercent));
             consumedBytes += lengthTillPercent;
             int addition = 1;
+            bool fillLeadingZeroes = false;
```
ForNeVeR, line 133: _(Minor)_ The CI correctly notes that this variable is only set and never used.

ForNeVeR:
This wasn't an issue of locale, but instead the formatting for `UIntPtr` was broken on .NET Framework: the behavior of `string.Format("{0:x2}", (IntPtr)0x20)` is different between net48 and net7.

I've fixed that with more portable code.

Good thing we have the integration tests :)

ForNeVeR (approved):
Thanks!

Issue #453, opened 2023-10-21
Local struct declaration can't be resolved


abrahamFerga:
Local struct declaration can't be resolved:
```
int main() {
    struct bar {
        int x;
        int y;
    };
    struct bar foo;
    foo.x = 10;
    return foo.x;
}
```

It works if the struct is declared globally.
Found this behavior while working in #450, it makes the following case fail:

```
int main() {
    struct bar {
        int x;
    };
    return sizeof(struct bar);
}
```

abrahamFerga:
Hello everyone, 

I could use some guidance regarding a topic in the C standard (6.2.1 Scopes of identifiers):

> `An identifier can denote an object; a function; a tag or a member of a structure, union, or enumeration;
a typedef name; a label name; a macro name; or a macro parameter. The same identifier can denote
different entities at different points in the program.`

My initial idea was to create a list of structures within the `GlobalConstructorScope`, `FunctionScope`, and `BlockScope`, similar to what we're doing for variables.

`private readonly Dictionary<string, VariableInfo> _variables = new();`

However, if we go this route, we might need to do the same for functions, structs, enums, typedefs, and more. What are your thoughts on this? Am I deviating from the main goal of Milestone 61? While it is possible to declare functions, structs, etc., locally, it's not as common, as far as I know.

```
int main() {
    int localFunc()
    {
        return 1;
    }
    return localFunc();
}
```

Cheers.

ForNeVeR:
I think that we should port the same structure we currently have for types in the `TranslationUnitContext`:
```
    private readonly Dictionary<IGeneratedType, TypeReference> _generatedTypes = new();
    private readonly Dictionary<string, IType> _types = new();
    private readonly Dictionary<string, IType> _tags = new();
```

Together with the accompanying methods (perhaps `ResolveType` as well?).

Perhaps let's try to extract it into a separate type, like `TypeDefinitionScope`, and then use it in `TranslationUnitContext` and other contexts that may own the type declarations?

> Am I deviating from the main goal of Milestone 61?

Meh, the milestones are here to somewhat direct the development, and definitely not to stop us from implementing whatever we like to :)

> While it is possible to declare functions, structs, etc., locally, it's not as common, as far as I know.

The most interesting features to implement are often the least common ones :)

PR #454, opened 2023-10-21
450 sizeof() operator

### LINKED ISSUE #450: Add support for "sizeof()" operator
The parser is currently failing to handle the "sizeof()" operator correctly. Here's an example that illustrates the issue:

`int main() {
    return sizeof(int);
}`

From the standard is Section 6.5.3.4 The sizeof and _Alignof operators
I will be working on this issue.
Related to #207

abrahamFerga:
This pull request closes #450, there are a few cases where sizeof() is failing (see the Integration Tests).

Let me know if any change is required or if I should wait for #453 (next on my list).


Cesium.CodeGen/Contexts/TranslationUnitContext.cs
```diff
@@ -168,7 +168,8 @@ internal IType ResolveType(IType type)
         return type;
     }
 
-    internal TypeReference? GetTypeReference(IGeneratedType type) => _generatedTypes.GetValueOrDefault(type);
+    internal TypeReference? GetTypeReference(IGeneratedType type) => _generatedTypes.GetValueOrDefault(type) ??
+                                                                     _generatedTypes.FirstOrDefault(t => t.Key.Identifier == type.Identifier).Value;
```
ForNeVeR: I don't believe this change is correct. The chance to mess something up (like `struct foo` vs `typedef foo struct { … }`) is very high.

I think that the type _should_ be emitted at this point.

For now, let's roll this statement back and fall back to `typedef` in a test, and figure the rest out in #453.

Cesium.CodeGen/Ir/Expressions/SizeOfOperatorExpression.cs
```diff
@@ -0,0 +1,54 @@
...
+            ({ }, { }) => LocalDeclarationInfo.Of(specifiers, abstractDeclarator).Type,
+            ({ }, null) => LocalDeclarationInfo.Of(specifiers, (Declarator?)null).Type
+        };
+    }
+
+    public IExpression Lower(IDeclarationScope scope)
+    {
+        if (_type is NamedType namedType)
+        {
+            var typeResolved = new IdentifierExpression(namedType.TypeName).Resolve(scope).GetValueType();
```
ForNeVeR, line 32: Currently, this definition, and perhaps even parsing for this, is incorrect.

We want to support both `sizeof(variable)` and `sizeof(type)`, right?

But so far the parser only supports `sizeof(type_name)`, and thus it _accidentally_ supports _variables_ that _look like type names_, which cannot be right 😅 

This leads to further problems, one of which is now demonstrated in a failed `typedef` test I pushed into the branch. In my opinion, it looks like you are trying to save the incorrect definition by resolving a `NamedType` as a variable name.

This is of course move in the right direction, but let's work a little bit more on the definition of this: perhaps let's introduce two different `sizeof` variants for types and variables? (and then accept just an `identifier` and see how it resolves in the lower/emit time)

Note I'm okay to disable the (currently) failing tests and merge this while opening a follow-up issue on this, but probably I'd like to disable the support for _variables_ completely beforehand. I am open for discussion, though. If you are very eager to merge this as is, I'm ready to negotiate somewhat :)

ForNeVeR (changes requested):

Cesium.CodeGen/Contexts/TranslationUnitContext.cs
ForNeVeR: Note I have rolled it back and pushed a commit to demonstrate an issue with this approach. Please tell if you are uncomfortable with such a move from me :)

abrahamFerga: I knew something was wrong with this but it was the only way I found to get the type. I will work in #453 and we won't need this anymore.
No worries, I'm here to learn, any feedback is well appreciated.

Cesium.CodeGen/Ir/Expressions/SizeOfOperatorExpression.cs
abrahamFerga: At first I had two different expressions for this but somehow it worked having only one of them as `sizeof(type_name)`, sure I will split both scenarios and I will update this pull request along the other issue mentioned above.

Thank you for the feedback!

abrahamFerga:
Hey there,

I've done some work to split up the cases for sizeof(Identifier) and sizeof(TypeName). It's working well in most situations, but there's a hiccup with structs since we haven't figured out a way to resolve them by identifier yet.

I'll be honest, my knowledge of C is somewhat limited, and the deeper I dive into it, the more peculiar things get. There are a few edge cases that this PR doesn't handle, like sizeof(5*5). However, I think these changes, while interesting, might not be necessary for the foreseeable future.

Cheers.

kant2002:
I personally believe that if this is half-done feature which does not cover all cases that's great addition. We work in small steps on such things. Because usually it's too draining work on full spec. If you think that you never introduce any "wrong" behaviour, then this should be fine.

Also, you doing great job. I really like to see all these additions to Cesium! Keep going!

ForNeVeR (approved):
Thanks!

PR #456, opened 2023-10-28
Removed unused namespaces

Lehonti:
Along the way, in the files that were modified by this change, I sealed `internal` classes and simplified `new` expressions


ForNeVeR (approved):
Thanks!

PR #457, opened 2023-10-28
Removed nesting levels in `CompoundInitializationExpression`

Lehonti:
Through `if` inversion and block-scoped `using`s


ForNeVeR (approved):
Thanks!

PR #458, opened 2023-10-28
`sealed` internal classes

Lehonti:
Along the way, some small improvements were also made, like turning some `switch` statements into `switch` expressions, and simplifying calls to constructors


ForNeVeR (approved):
Thanks!

PR #464, opened 2023-10-31
Used null-coalescing operator for nullable assignments

Lehonti:
With the added benefit that some `int?` variables now became `int`


ForNeVeR (approved):
Thanks!

PR #466, opened 2023-11-11
Add missing validation results

kant2002:
Something fishy is going on.


ForNeVeR:
Now wait a minute. How come the tests are green without these files?

ForNeVeR:
Found it. The corresponding tests are defined to return `void` instead of `Task`s.

ForNeVeR (approved):
Thanks for noticing!

PR #468, opened 2023-11-11
use STJ demo

Ilchert:


ForNeVeR:
Thanks! That demonstrates a possible approach well.

While the solution is… let's say _tolerable_, so far there are no plans to use it.

Some downsides (which are not blockers for this project, but here I am thinking of general critic of STJ) are:
- it seem to break if the derived types are in another assembly
- if does weird logic w.r.t. sealed/non-sealed types (perhaps it's possible to fix? but we don't care that much I'd say)
- we have to apply arbitrary type filters in the `Modifier` method, and, to my understanding, that's not a general solution, but a solution for particular case of our own assembly. If we ever move the types to a new namespace, or want to use types from another namespace (and I think we already want to serialize things from outside of `Cesium.Ast`), then we have to update the serialization configuration. That troubles me.

PR #470, opened 2023-11-12
Allow perform indexing over result of indexing

kant2002:
Example of added pattern
```
argc[i][t]
```

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/Values/LValueArrayElement.cs
+++ b/Cesium.CodeGen/Ir/Expressions/Values/LValueArrayElement.cs
@@ -2,6 +2,7 @@ using Cesium.CodeGen.Contexts;
 using Cesium.CodeGen.Ir.Types;
 using Cesium.Core;
 using Mono.Cecil.Cil;
+using System.Security.AccessControl;
 
 namespace Cesium.CodeGen.Ir.Expressions.Values;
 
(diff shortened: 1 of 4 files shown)


Cesium.CodeGen/Ir/Expressions/Values/LValueArrayElement.cs
```diff
@@ -59,13 +59,34 @@ private void EmitPointerMoveToElement(IEmitScope scope)
...
+            PointerType pointerType => pointerType.Base,
+            _ => throw new AssertException("Array or pointer type expected.")
+        };
+        if (baseType is InPlaceArrayType or PointerType)
+        {
+            return GetBaseType(baseType);
+        }
+
+        return baseType;
+    }
```
ForNeVeR, line 76: I feel like this all is broken.

`LValueArrayElement.GetValueType()` should return the type of the array element, while it currently returns the array type itself.

While this `GetBaseType()` should only perform one step of indirection and not be recursive!

In your PR and even before that, it seems to always remove all the indirection levels and I don't understand this behavior at all.

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Expressions/Values/LValueArrayElement.cs
kant2002: You know, you are right. Thanks your for keeping me in check.

ForNeVeR (approved):
Thanks!

PR #471, opened 2023-11-12
Add getenv and strchr

kant2002:
### After merge
- [ ] If we've decided to leave the current implementation of `getenv`, then open a new issue on the improvements (see the comments for this PR for details)

### DIFF
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -286,7 +286,7 @@ public unsafe static class StdIoFunctions
         }
 
         var bytes = encoding.GetBytes(str);
-        var storage = (byte*)StdLibFunctions.Malloc((nuint)bytes.Length);
+        var storage = (byte*)StdLibFunctions.Malloc((nuint)bytes.Length + 1);
         for (var i = 0; i < bytes.Length;i++)
         {
             storage[i] = bytes[i];
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -1,7 +1,5 @@
 using System.Runtime.InteropServices;
-#if NETSTANDARD
 using System.Text;
-#endif
 
 namespace Cesium.Runtime;
 
@@ -244,6 +242,24 @@ public unsafe static class StdIoFunctions
 #else
         return Marshal.PtrToStringUTF8((nint)str);
 #endif
+    }
+
+    internal static byte* MarshalStr(string? str)
+    {
+        Encoding encoding = Encoding.UTF8;
+        if (str is null)
+        {
+            return null;
+        }
+
+        var bytes = encoding.GetBytes(str);
+        var storage = (byte*)StdLibFunctions.Malloc((nuint)bytes.Length);
+        for (var i = 0; i < bytes.Length;i++)
+        {
+            storage[i] = bytes[i];
+        }
+
+        return storage;
     }
 
     private static StreamHandle? GetStreamHandle(void* stream)
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -292,6 +292,7 @@ public unsafe static class StdIoFunctions
             storage[i] = bytes[i];
         }
 
+        storage[bytes.Length] = 0;
         return storage;
     }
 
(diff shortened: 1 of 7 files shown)


kant2002:
This is interesting case for me with `getenv`. It seems that we need to populate some internal structure in our entry point for environment and then operate on that local data. I prefer somehow avoid that for foreseeable future. Maybe you can guess what kind of post processing needed for Lunix/Mac?

ForNeVeR:
Linux tests fail in interesting manner:
```
Expected: ···ocal/games:/snap/bin\n
Actual:   ···ocal/games:/snap/bin�t4�\n
```

Will you take a look at this?

ForNeVeR:
> This is interesting case for me with `getenv`. It seems that we need to populate some internal structure in our entry point for environment and then operate on that local data. I prefer somehow avoid that for foreseeable future. Maybe you can guess what kind of post processing needed for Lunix/Mac?

Ok, I see the issue here. The current implementation allocates a new string on each call, while the C standard implies that we reuse the same memory block for the program environment.

I believe it's okay to leave the current implementation you wrote for now, but open a new issue for its improvement.

The _improvement_ in question, I believe, would be to:
- on the first call to `getenv`, allocate a big memory block for all the environment strings, and remember the indices for each variable somewhere (i.e. in `Dictionary<string, int>`)
- put all the variables into the block and separate with `\0`
- terminate with `\0\0`

This will more or less mimic how glibc works, prevent any leaks, and avoid unnecessary allocations. We may need to reconsider how it all works in the future (in particular, when to invalidate the buffer returned by `getenv()`), if we decide to add functions that modify the environment variables (interestingly, they aren't included in the standard, even though they are in POSIX and alike, so we'll likely want to add them eventually anyway).

Also, it will open the road for alternate POSIX-compatible `main` signature (i.e. the `main(argc, argv, envp)`).

The standard allows us to invalidate the buffer after the next call to `getenv()`, though I think there may be programs that rely on it not being invalidated.

Cesium.Runtime/StdLibFunctions.cs
```diff
@@ -138,6 +138,17 @@ public static int Atoi(byte* ptr)
...
 
+    public static byte* GetEnv(byte* ptr)
+    {
+        var str = StdIoFunctions.Unmarshal(ptr);
+        if (str is null)
+        {
+            return null;
+        }
+
+        return StdIoFunctions.MarshalStr(Environment.GetEnvironmentVariable(str)?.TrimEnd('\\'));
```
ForNeVeR, line 149: Why `TrimEnd('\\')`? I don't think it's normal for a runtime environment to strip trailing slashes from a variable like this.

ForNeVeR (changes requested):

Cesium.Runtime/StdLibFunctions.cs
kant2002: Similar to https://github.com/ForNeVeR/Cesium/pull/471#issuecomment-1807255956 I cannot understand why there different values in C / .NET, Maybe I should try to use other API? anyway I did not have any reasonable ideas, and put this "bad code" to show you the issues. Will try take a look.

ForNeVeR: The API you use looks correct. What's the particular difference you observe? Or do you want me to take a look myself?

kant2002: I see one `\` at the end of PATH variable. Also on Linux we see `�t4��`. Maybe that's artifact of `printf`, don't know

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -279,6 +277,24 @@ public static int FPrintF(void* stream, byte* str, void* varargs)
...
+    internal static byte* MarshalStr(string? str)
+    {
+        Encoding encoding = Encoding.UTF8;
+        if (str is null)
+        {
+            return null;
+        }
+
+        var bytes = encoding.GetBytes(str);
+        var storage = (byte*)StdLibFunctions.Malloc((nuint)bytes.Length);
```
ForNeVeR, line 289: Gotcha!
```suggestion
        var storage = (byte*)StdLibFunctions.Malloc((nuint)bytes.Length + 1);
```

You forgot to allocate space for the terminating null character.

Cesium.Runtime/StdLibFunctions.cs
ForNeVeR: Regarding `\`, I believe it is normal: `PATH` variable (especially on Windows) in particular may just end with `\`, this is just its value.

ForNeVeR (approved):
Thanks!

PR #480, opened 2023-11-17
Address perf+correctness of StrLen and StrChr

neon-sunset:


Cesium.Runtime/StringFunctions.cs
```diff
@@ -14,26 +16,8 @@ public unsafe static class StringFunctions
...
-        Encoding encoding = Encoding.UTF8;
-        int byteLength = 0;
-        byte* search = str;
-        while (*search != '\0')
-        {
-            byteLength++;
-            search++;
-        }
-
-        int stringLength = encoding.GetCharCount(str, byteLength);
```
neon-sunset, line 32: This could produce incorrect behavior - `strlen` is index of \0, GetCharCount calculates the length in UTF-16 code units, which will differ should the bytestring contain non-ASCII chars.

Cesium.Runtime/StringFunctions.cs
```diff
@@ -14,26 +16,8 @@ public unsafe static class StringFunctions
...
-        return (uint)stringLength;
-#else
-        return (uint)(Marshal.PtrToStringUTF8((nint)str)?.Length ?? 0);
```
neon-sunset, line 35: Transcoding to UTF-16 string is quite wasteful for finding a null byte offset and would not handle lengths [greater than ~2GB](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/String.cs#L39C40-L39C50).

Cesium.Runtime/StringFunctions.cs
```diff
@@ -186,19 +170,21 @@ public static int StrNCmp(byte* lhs, byte* rhs, nuint count)
...
-        }
+            int match;
+            nuint offset = 0;
+            byte c = (byte)ch;
 
-        while (*str != 0)
-        {
-            if (*str == ch)
+            // TODO: Copy IndexOfNullByte impl. from CoreLib in a distant future
+            var span = new Span<byte>(str, int.MaxValue);
```
neon-sunset, line 180: A no-op once compiled (span is optimized away), we have to do this sadly because `CoreLib` keeps all the nice methods for working with raw ptrs internal.

neon-sunset:
Aha, we get AVE. Let me check.

ForNeVeR (commented):
I'll leave a comment while there's still ongoing work.

I agree that the new approach is more correct (your points on UTF-8 and UTF-16 are all valid), and so yeah, I'd like to see the tests fixed and this merged.

I am also open for discussion on leaving the netstandard and only supporting .NET 7 (or even 8) in the stdlib, if it becomes too painful for us to properly support the older target frameworks.

neon-sunset:
@ForNeVeR In this particular case it's not an issue at all, to support NS2.0. It might, however, become an issue in the future.

A few thoughts: I don't think trying to reimplement stdlib is a scalable approach - it will either take too much effort or the quality of the resulting code may be insufficient.

Ideally, we should be mixing and matching the following options on a case-by-case basis:
- Forwarding calls to .NET CoreLib
- P/Invoking host-provided stdlib, like .NET itself does (e.g. it will call `memcpy` above certain threshold, even though today its own implementation literally does a better job than most due to conservative compilation flags of your average glibc)
- Handling happy-path in C# and then deferring to either of the above

P/Invokes are already very cheap/almost free in C# (Жова and Go could never) but in the future, if it ever becomes a concern, they can be made as cheap as calls in an actual compiled C code by applying `[SuppressGCTransition]`, generating `DirectPInvoke`s for NativeAOT and caching function pointers to a readonly static for JIT (if it doesn't do it already).

Cesium.Runtime/StringFunctions.cs
```diff
@@ -14,26 +7,22 @@ public unsafe static class StringFunctions
...
-        Encoding encoding = Encoding.UTF8;
-        int byteLength = 0;
-        byte* search = str;
-        while (*search != '\0')
-        {
-            byteLength++;
-            search++;
+            // TODO: Copy IndexOfNullByte impl. from CoreLib in a distant future
+            while ((match = new Span<byte>(str + offset, int.MaxValue)
+                .IndexOf((byte)0)) < 0)
```
ForNeVeR, line 17: While I value the performance of these vectorized functions, the use of `int.MaxValue` here seems hazardous.

In particular, there seems to not be any guarantee of `Span.IndexOfAny` not accessing the bytes after the needle is found. See the implementation: it will happily load data in chunks of 8 or 4 bytes.

If said data lies at the end of the memory page, this will easily trigger a memory access violation.

I was able to reproduce some sort of runtime crash in the following test:
```csharp
[Fact]
public void StrLen2()
{
    for (var i = 1_000_000; i < 2_000_000; ++i)
    {
        var init = new string('x', i);
        var bytes = Encoding.UTF8.GetBytes(init).Append((byte)0).ToArray();
        fixed (byte* str = bytes)
        {
            var actual = StringFunctions.StrLen(str);

            Assert.Equal((nuint)i, actual);
        }
    }
}
```

Not sure how to proceed from here, though. What do you think? To me, it looks like it's _impossible in principle_ to properly determine the string length in a vectorized way, because we don't know how long ahead we are allowed to read.

We can perform this in `StrChr` if we determine `StrLen` first, and then perform the vectorized search. But I'm not sure how efficient it will be practically. We can try to benchmark this, I guess.

But I think I'd prefer a dumb but correct implementation first.

ForNeVeR (changes requested):

Cesium.Runtime/StringFunctions.cs
neon-sunset: > Not sure how to proceed from here, though. What do you think? To me, it looks like it's impossible in principle to properly determine the string length in a vectorized way, because we don't know how long ahead we are allowed to read.

We have three options here:
- give up and revert
- calculate the distance from the pointer to memory page end (cheap), and then scan in pagesize-long chunks
- properly adapt IndexOf... code from SpanHelpers in CoreLib and write a vectorized routine here that would do the right thing and won't spend time/overhead on calling IndexOf (I'm yet to check what other strchr implementations do and if they are not vectorized manually but auto-vectorized by compiler, what the asm looks like)

I'm gonna give it a try and if it's too much code, will put it in a gist or comment out and revert until Cesium is more feature-complete :)

kant2002: Hey :) how it can become more feature complete (and what this is mean in this context?) if people like you back off of improvements ;)

also if you want very specific feature from Cesium, please tell. We would try to make it happens

ForNeVeR: > calculate the page boundary offset, scan once with `new Span<byte>(ptr, pageEnd)`, and then scan in pagesize-long chunks

I am against the page-boundary shenanigans, just because there are no guarantees of memory pages even existing in .NET runtime. How will WASM or an alternate runtime work with them? No idea. So, we'll need a fallback solution for this case, and if we already have a fallback, then I'm okay with just using it always, to reduce testing surface.

> properly adapt IndexOf... code from SpanHelpers in CoreLib and write a vectorized routine here that would do the right thing

My point is that "the right thing" is _impossible_ to write in a vectorized way _before_ determining the boundaries of the memory region we operate on. So, vectorized `strlen` is impossible.

(Also I wonder how's it done in other C implementations.)

neon-sunset: > (Also I wonder how's it done in other C implementations.)

Those do it with aligning the pointer and then doing register-sized loads and bitmasks: https://elixir.bootlin.com/glibc/glibc-2.17/source/string/strlen.c

I assume page-boundary shenanigans are easily done with calculating the distance from the pointer to the next 4KiB-boundary or whatever pagesize is if we get it from the system.

ForNeVeR: For Cesium, .NET is the target platform, and ideally it should not leak nor rely on any underlying details.

Of course, this mindset may not always be practical. But for this particular case, I believe it is still.

neon-sunset: We're relying on the details of .NET itself which has a few expectations from the platforms it supports like requiring support for unaligned reads/writes, specific FP math and its exceptions handling, etc. As long as we fulfill the behavior contract - it can be considered a correct implementation.

I think there is little profit in debating this at the current stage of development of Cesium and can be revisited later.

neon-sunset:
@ForNeVeR The issue with SEGFAULT should be addressed now - I opted into a simpler approach with aligning the pointer at 16B boundary first and then using 128b vectors for searching the matches directly. This way it should never cross a page boundary (which was causing issues).

neon-sunset:
I had to bump up `.Runtime` to `net7.0` because cross-platform Vector API is not available on older targets.

Cesium.Runtime.Tests/Cesium.Runtime.Tests.csproj
```diff
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <TargetFrameworks>net7.0</TargetFrameworks>
-        <TargetFrameworks Condition=" $([MSBuild]::IsOsPlatform('Windows')) ">$(TargetFrameworks);net48</TargetFrameworks>
+        <!-- <TargetFrameworks Condition=" $([MSBuild]::IsOsPlatform('Windows')) ">$(TargetFrameworks);net48</TargetFrameworks> -->
```
ForNeVeR, line 5: After some thought, I still want us to support .NET Framework on Windows. Let's re-enable that here, and add a fallback implementation for .NET Standard.

Cesium.Runtime.Tests/StringFunctionTests.cs
```diff
@@ -0,0 +1,75 @@
+using System.Text;
+
+namespace Cesium.Runtime.Tests;
+
+public unsafe class StringFunctionTests
```
ForNeVeR, line 5: Let's add some Unicode tests, shall we?

Cesium.Runtime/Cesium.Runtime.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFramework>net7.0</TargetFramework>
```
ForNeVeR, line 4: I would still insist on having netstandard2.0 support via `#if`.

Cesium.Runtime/StringFunctions.cs
```diff
@@ -14,50 +12,57 @@ public static unsafe class StringFunctions
...
     public static nuint StrLen(byte* str)
     {
-#if NETSTANDARD
-        if (str == null)
+        if (str != null)
         {
-            return 0;
-        }
+            var start = str;
+            while ((nuint)str % 16 != 0)
```
ForNeVeR, line 18: I suggest introducing a constant for this `16`, like `MinimalPageBoundary` or something, to make the size choice more clear.

ForNeVeR (changes requested):
I am ok with the implementation, but would insist on having a fallback for older runtimes. Do you want to implement it yourself, or will entrust it to me?

Cesium.Runtime/StringFunctions.cs
ForNeVeR: Verified the aforementioned test on the new code, and it works now.

Cesium.Runtime/Cesium.Runtime.csproj
neon-sunset: Hmm, this is runtime. Is it planned to be referenced as a package by .NET Framework/pre-.NET 7 targets or is it packaged as a standalone, well, runtime?

Cesium.Runtime/StringFunctions.cs
neon-sunset: Right, it's just `Vector128<byte>.Count` which I omitted because surrounding code is compact too, let me fix it. (to which we align - this is both for performance and for never crossing page boundary, whatever size it is, it occured to me we don't need actual page size here since it's always bigger than 16B and 16B alignment will never cross it)

neon-sunset:
> I am ok with the implementation, but would insist on having a fallback for older runtimes. Do you want to implement it yourself, or will entrust it to me?

Alright, I will if-def it. "Proper" way to fix it is by copying impl. from Glibc with its bitwise tricks (which we, in .NET, do not need) but I guess "works on NETFW, if you care about performance that's on you" is also fine.

Cesium.Runtime.Tests/StringFunctionTests.cs
neon-sunset: Sure, will do

Cesium.Runtime/Cesium.Runtime.csproj
ForNeVeR: Yes, it is planned to be referenced by projects targeting older .NET versions.

PR #481, opened 2023-11-18
Fix codegen with forward declaration.

kant2002:
If forward declaration has different parameter name then code gen does not working


ForNeVeR (approved):
Thanks!

PR #482, opened 2023-11-18
Add tests for compilation of two source files with identical types.

kant2002:


kant2002:
Probably depends on #481 since I was testing from that branch.

Cesium.CodeGen/Ir/Types/EnumType.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Cesium.CodeGen.Ir.Types;
 
-internal sealed class EnumType : IType
+internal sealed class EnumType : IType, IEquatable<EnumType>
```
ForNeVeR, line 7: The compiler warnings here are correct, let's implement `GetHashCode` as well 😅

Cesium.CodeGen/Ir/Types/StructType.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Cesium.CodeGen.Ir.Types;
 
-internal sealed class StructType : IGeneratedType
+internal sealed class StructType : IGeneratedType, IEquatable<StructType>
```
ForNeVeR, line 10: Let's implement `GetHashCode` here, too.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #484, opened 2023-11-19
Add support for variadic functions

kant2002:


Cesium.CodeGen/Contexts/FunctionScope.cs
```diff
@@ -62,7 +62,17 @@ public VariableDefinition ResolveVariable(string identifier)
 
         return variableDefinition;
     }
-    public ParameterInfo? GetParameterInfo(string name) => FunctionInfo.Parameters?.Parameters.FirstOrDefault(p => p.Name == name);
+    public ParameterInfo? GetParameterInfo(string name)
+    {
+        var parametersInfo = FunctionInfo.Parameters;
+        if (parametersInfo is null) return null;
+        if (name == "varargs" && parametersInfo.IsVarArg)
```
ForNeVeR, line 69: Let's call it `__varargs` to not clash with any user-defined names.

Cesium.CodeGen/Ir/Types/PrimitiveType.cs
```diff
@@ -149,6 +149,7 @@ Kind switch
                 TargetArchitectureSet.Bit64 => 8,
                 _ => throw new AssertException($"Architecture set not supported: {arch}.")
             },
+            PrimitiveTypeKind.Void => 1,
```
ForNeVeR, line 152: Welp, this one is weird. Technically we shouldn't define `sizeof(void)`. Why is that required?

Cesium.CodeGen.Tests/CodeGenMethodTests.cs
```diff
@@ -361,10 +361,11 @@ int main()
 }");
 
     [Fact]
-    public void InvalidPointerWithIntSubtractionTest() => DoesNotCompile(@"int main() {
+    public void InvalidPointerWithIntSubtractionTest() => DoTest(@"int main() {
```
ForNeVeR, line 364: _(Minor)_ The name of this test is now misleading: it is no longer `Invalid`, right?

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Types/PrimitiveType.cs
kant2002:
```
void* ptr;
ptr += 8;
```

ForNeVeR: Oh wow, GCC also thinks `sizeof(void)` is 1. Okay.

Cesium.Compiler/stdlib/stdarg.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+typedef void* va_list;
+
+#define va_start(list,...) list = __varargs
+#define va_copy(dest, src) dest = src;
```
ForNeVeR, line 6: I've decided to remove `;` from here, to force the user to put `;` on their own.

ForNeVeR (approved):
Thanks!

ForNeVeR:
Whoops, it looks like we have a conflict with your own changes from #481 😅 

Could you please take a look?

ForNeVeR (approved):
Thanks!

PR #499, opened 2023-12-04
Subscripts of complex expressions

### LINKED ISSUE #230: Subscript expressions (aka index operator) of complex expressions

### LINKED ISSUE #477: `LValueArrayElement::GetValueType` is implemented weirdly
Currently, `LValueArrayElement::GetValueType` returns the type of the array itself. I don't think it is a correct implementation: it should actually return the type of the array element, not the array itself.

But we'll also need to figure out how and why it works currently with an incorrect implementation. How's that `GetValueType` used anyway?

ForNeVeR:
- Closes #230.
- Closes #252 (supersedes that PR as it includes the changes from it).
- Closes #477.

## TODO
- [x] Figure out what's a "complex expression" exactly
- [x] Fix the issue
- [x] Green tests


ForNeVeR:
Was closed accidentally. I am still working on it.

Issue #489, opened 2023-11-23
Make `FunctionInfo` immutable


ForNeVeR:
The current form of `Cesium.CodeGen.Contexts.Meta.FunctionInfo` is confusing: it is an immutable record struct, but with several mutable fields.

Let's make it fully immutable.

While we are at it, remove `TODO[#489]` from the code.

PR #500, opened 2023-12-08
Resolved Issue #489 : Made FunctionInfo immutable

lightlessdays:
Changes made:
[x] Removed the setters from properties like Parameters, StorageClass, and IsDefined.
[x] Initialized the CliImportMember property using init instead of in the constructor. This makes it immutable after construction.
[x] Remove TODO[#489] from the code.

Signed-off-by: @lightlessdays

### DIFF
--- a/Cesium.CodeGen/Contexts/Meta/FunctionInfo.cs
+++ b/Cesium.CodeGen/Contexts/Meta/FunctionInfo.cs
@@ -5,55 +5,53 @@ using Cesium.CodeGen.Ir.Types;
 using Cesium.Core;
 using Mono.Cecil;
 
-namespace Cesium.CodeGen.Contexts.Meta;
-
-// TODO[#489]: This is confusing, make immutable.
-internal record FunctionInfo(
-    ParametersInfo? Parameters,
-    IType ReturnType,
-    StorageClass StorageClass,
-    bool IsDefined,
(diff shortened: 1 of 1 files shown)


lightlessdays (commented):
Haven't changed many lines, yet Git shows that I have changed 80 lines. This is due to the reason that I deleted the TODO comment, which made all the code shift one line up.

ForNeVeR:
Actually, the change you did to cause 80 changed lines is to add an old-school namespace. Before it was
```csharp
namespace Cesium.CodeGen.Contexts.Meta;

internal record FunctionInfo(…
```

And now it is
```csharp
namespace Cesium.CodeGen.Contexts.Meta
{
    internal record FunctionInfo(…
```

I am not too picky on the code style and will review the change in detail anyway, but I would say that it's better to turn the namespace declaration back to how it was written. But I can roll it back myself when I'll merge the change, so it's not a big deal anyway.

ForNeVeR:
The unit tests are broken, though. Could you please take a look?

Of course, I can help if you need.

ForNeVeR:
Of course, feel free to update the code to make it compile after the refactoring :)

> Initialized the CliImportMember property using init instead of in the constructor. This makes it immutable after construction.

I wonder why you moved other properties to the constructor, but left this one in the class block. Is there a particular reason for that I'm missing?

PR #501, opened 2023-12-09
Allow typdefs on partial declarations for struct/enum

kant2002:
disallow enum and struct to share name


Cesium.CodeGen/Contexts/TranslationUnitContext.cs
```diff
@@ -115,7 +119,26 @@ internal void AddTypeDefinition(string name, IType type)
...
+        }
+
+        _tags.Add(name, type);
+
+        string GetTypeKind(Type type) => type.Name switch
+        {
+            nameof(StructType) => "struct",
+            nameof(EnumType) => "enum",
+            _ => throw new InvalidOperationException($"Unsupported type {type} used."),
+        };
```
ForNeVeR, line 140: This fragment deserves some more attention.

1. Let's get rid of comparing `GetType()` directly.
2. Also, let's get rid of block like `GetTypeKind(Type type)` (I don't like that the `Type` gets passed around).

Instead, I'd prefer an `enum TypeKind` and a corresponding property in the `IType` interface.

Cesium.CodeGen/Contexts/TranslationUnitContext.cs
```diff
@@ -160,6 +183,18 @@ internal IType ResolveType(IType type)
...
+            {
+                if (_types.TryGetValue(structType.Identifier, out var existingType))
+                {
+                    if (existingType is StructType existingStructType && existingStructType.Members.Count == 0)
+                    {
+                        existingStructType.Members = members;
+                        return existingType;
+                    }
+                }
+            }
```
ForNeVeR, line 196: I'm leaving a note for myself (open for discussion, of course): we seem to use an empty struct as a placeholder for a forward-declared struct. I think this could be improved: let's add a special mark for a forward-declared struct explicitly.

This may become a problem if we ever start supporting empty structs (as a potential extension, though I can't remember if they probably already made it to the standard).

Cesium.CodeGen/Ir/Types/PointerType.cs
```diff
@@ -50,4 +50,8 @@ public IExpression GetSizeInBytesExpression(TargetArchitectureSet arch)
 
         return new SizeOfOperatorExpression(this);
     }
+    public override int GetHashCode()
+    {
+        return base.GetHashCode();
+    }
```
ForNeVeR, line 56: Oh, is it really necessary? :)

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Types/PointerType.cs
kant2002: This is requred to break cycle in struct GetHashCode when struct has pointer of itself's as member type.

Cesium.CodeGen/Contexts/TranslationUnitContext.cs
kant2002: Based on what I see during implementation, and how we really construct type information.
It seems to me that instead of bunch of clases we can have 1 (probably) structure which repesents whole set of types.
```csharp
struct TypeInformation
{
    string name;
    TYpeKind Kind;
    bool IsPointer;
    TypeInformation * baseType;
    bool IsConstant;
    int Size;
    // Maybe some other properties.
}
```

That allow get rid of some type gymnastics which I have to do in some places already

Cesium.CodeGen/Ir/Types/PointerType.cs
ForNeVeR: Oh boy

Cesium.CodeGen/Contexts/TranslationUnitContext.cs
ForNeVeR: This would probably also be okay, though I'd prefer to review that as a separate PR.

Cesium.CodeGen/Ir/Types/PointerType.cs
ForNeVeR: Perhaps let's break it in the struct type then somehow? It looks weird to break the `PointerType` because of that. In its current state, it is not very useful as a key in a hash table.

Cesium.Compiler/Properties/launchSettings.json
```diff
@@ -0,0 +1,104 @@
+{
+    "profiles": {
+        "Cesium.Compiler": {
+            "commandName": "Project",
+            "commandLineArgs": "C:\\d\\github\\Cesium\\Cesium.IntegrationTests\\variable_storage.c -o C:\\d\\github\\Cesium\\Cesium.IntegrationTests/bin/out_cs.exe"
```
ForNeVeR, line 5: I am not against this file, but let's replace the paths with relative ones :)

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

Cesium.CodeGen/Contexts/TranslationUnitContext.cs
ForNeVeR: See #504.

PR #505, opened 2023-12-24
Add bunch of functions

kant2002:


kant2002:
I would like to present case when we have incompatibility with standard(or maybe most common implementations) regarding `long`, `unsigned long`

ForNeVeR:
Tests have failed, could you please take a look?

> I would like to present case when we have incompatibility with standard(or maybe most common implementations) regarding `long`, `unsigned long`

Could you please clarify what particular incompatibility you mean?

In my view, we are
- 100% compatible with standard w.r.t. type ranges and possible values
- compatible with the so-called [LP64 model](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models), mostly used in Linux and portable C implementations (but notably not MSVC)

Whether we count LP64 as "the most common" or not is open for discussion (I would say _yes_), but I also believe it to be the most sane of the two, and that's the reason I chose it for Cesium.

kant2002:
Found couple stupid bugs in tests. Disable tests which are incompatible across platforms. not only negative values behave differently, but also out of range one. Windows most simplistic, and do not pedantic.

ForNeVeR (approved):
Thanks!

kant2002:
I am thinking how can we capture behavior of Cesium in cases where platform do not agree between themselves. Gold tests with Verify is too risky for now.

PR #506, opened 2023-12-24
Improve enum resolutions

kant2002:
Allow enums values to be used when defined via typedef
Allow passing enum values to functions


ForNeVeR (approved):
Thanks!

PR #507, opened 2023-12-24
Resolve sizeof on global struct definitions

### LINKED ISSUE #453: Local struct declaration can't be resolved

kant2002:
Fixes #453


ForNeVeR (approved):
Thanks!

PR #512, opened 2024-01-06
Add __FILE__ and __LINE__ support

kant2002:
Also double hash and consolidate dependencies


ForNeVeR:
As you can see in the integration test results, something's weird with how we count the lines 😅 

Could you please take a look at that?

ForNeVeR:
Windows test is still failing. Wanna take a look? Or do you need help?
image

ForNeVeR (approved):
Thanks!

Issue #61, opened 2022-01-16
Milestone: sha1collisiondetection


ForNeVeR:
Historically, the idea of Cesium appeared after we were discussing a possibility of porting code from [sha1collisiondetection](https://github.com/cr-marcstevens/sha1collisiondetection/) project to [Fenrir](https://github.com/ForNeVeR/Fenrir), a project to create a Git client in .NET.

So, our next milestone is going to be the compilation of said project using Cesium.

Milestone goals:
- take as much of the code as possible in its pristine state
- make sure the project compiles and works
- run project tests

Milestone non-goals:
- take **all** of the project in its pristine state
- integrate with any external build system (say, Make, CMake, MSBuild)

Tasks (a pretty rough list; will be elaborated on the go):
- [ ] Multiple translation unit support
- [ ] Function declaration support
- [ ] Preprocessor macro support (`#define`, `#ifdef`)
- [ ] Preprocessor interpreter language support (`#if defined`)
- [ ] Struct definitions (not a lot of them, thankfully)
- [ ] Codegen: `switch-case` support
- [ ] Codegen: global variables
- [ ] More of the standard library to implement
- [ ] More parser constructs

ForNeVeR:
Project fork for experiments with a rudimentary build script: https://github.com/ForNeVeR/sha1collisiondetection/blob/a00de8ddf23420057013a583551c9c31b48abe50/build.cesium.ps1

PR #519, opened 2024-01-20
7645re/added conditional elif block

### LINKED ISSUE #517: Preprocessor: support `#elif`
Example:
```c
#if FOO
// foo
#elif BAR
// bar
#endif
```

Currently, we don't support `#elif`.

7645re:
Closes #517.

this is still a draft, there are problems with naming, but I'm still more concerned about finding more corner cases to fully test the code for performance and optimization


ForNeVeR (approved):
So far looks good to me. Please let me know when you are finished with the naming stuff you mentioned.

kant2002:
Not important, but maybe `#elifdef` and `#elifndef` can be added here as well?

ForNeVeR:
I would suggest against extending the task definition. Let's focus on immediate issues and add other features later.

In particular, #61 is blocked on `#elif` specifically.

7645re:
Okay, let's finish this task and then create an issue marked area:standard-support?

ForNeVeR (approved):
Thanks!

PR #524, opened 2024-01-22
(#478) allocate once and reuse memory for env vars

### LINKED ISSUE #478: Get rid of memory leaks during `getenv`
Our current implementation of `getenv` leaks memory: it allocates blocks for strings, but never reclaims them.

From discussion of #471:

> The current implementation allocates a new string on each call, while the C standard implies that we reuse the same memory block for the program environment.
> 
> I believe it's okay to leave the current implementation you wrote for now, but open a new issue for its improvement.
> 
> _The improvement_ in question, I believe, would be to:
> - on the first call to `getenv`, allocate a big memory block for all the environment strings, and remember the indices for each variable somewhere (i.e. in `Dictionary<string, int>`)
> - put all the variables into the block and separate with `\0`
> - terminate with `\0\0`
> 
> This will more or less mimic how glibc works, prevent any leaks, and avoid unnecessary allocations. We may need to reconsider how it all works in the future (in particular, when to invalidate the buffer returned by `getenv()`), if we decide to add functions that modify the environment variables (interestingly, they aren't included in the standard, even though they are in POSIX and alike, so we'll likely want to add them eventually anyway).
> 
> Also, it will open the road for alternate POSIX-compatible `main` signature (i.e. the `main(argc, argv, envp)`).
> 
> The standard allows us to invalidate the buffer after the next call to `getenv()`, though I think there may be programs that rely on it not being invalidated.

KeterSCP:
Closes #478

### DIFF
--- a/Cesium.Runtime/StdLibFunctions.cs
+++ b/Cesium.Runtime/StdLibFunctions.cs
@@ -1,6 +1,8 @@
+using System.Collections;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Text;
 
 namespace Cesium.Runtime;
 
@@ -12,8 +14,22 @@ public unsafe static class StdLibFunctions
     public const int RAND_MAX = 0x7FFFFFFF;
     private static Random shared = new();
 
+    private class EnvVarsStorage
+    {
+        public byte* Values { get; }
+        public Dictionary<string, int> Indices { get; }
+
+        public EnvVarsStorage(int bufferLength, Dictionary<string, int> indices)
+        {
+            Values = (byte*)Malloc((nuint)bufferLength);
+            Indices = indices;
+        }
+    }
+
     [FixedAddressValueType]
     private static int errNo;
+
+    private static EnvVarsStorage? _envVarsStorage;
 
     public static int Abs(int value)
     {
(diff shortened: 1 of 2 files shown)


Cesium.Runtime/StdLibFunctions.cs
```diff
@@ -139,13 +155,21 @@ public static int Atoi(byte* ptr)
...
-        if (str is null)
+        var envKey = StdIoFunctions.Unmarshal(ptr);
+
+        if (envKey is null)
         {
             return null;
         }
 
-        return StdIoFunctions.MarshalStr(Environment.GetEnvironmentVariable(str));
+        _envVarsStorage ??= InitEnvVarsStorage();
```
ForNeVeR, line 165: I am a little concerned about the thread safety here. If two threads enter this field, then `_envVarsStorage` will just silently leak.

Let's just make it `Lazy`? What do you think?

ForNeVeR (changes requested):
Thanks, this is great! One minor issue though.

Cesium.Runtime/StdLibFunctions.cs
KeterSCP: Good point! Fixed it

ForNeVeR (approved):
Thanks!

PR #527, opened 2024-01-23
7645re/added conditional elifdef elifndef blocks

### LINKED ISSUE #520: Preprocessor: support `#elifdef` `#elifndef`
Example:

```c
#if FOO
// foo
#elifdef BAR
// bar
#elifndef BOB
// bob
#endif
```
Currently, we don't support #elifdef & #elifndef.

7645re:
Closes #520


ForNeVeR (approved):
Thanks!

PR #528, opened 2024-01-23
Preprocessor overhaul

### LINKED ISSUE #522: Preprocessor: A lot of unnecessary actions
Now our preprocessor goes deep into the conditional structure and processes instructions inside it, even if the conditional block has the false state, I think it needs to be redone so that it does not go deep into the conditional structure if it is false.

ForNeVeR:
Closes #522.

This migrates our preprocessor to a separate parser and AST model better suited for nested `#if` block processing.

## TODO
- [x] Green tests
- [x] TODO sweep
- [x] Review
- [ ] Merge
- [ ] Open new issues


Cesium.Preprocessor/Cesium.Preprocessor.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>net7.0</TargetFramework>
+        <TargetFramework>net8.0</TargetFramework>
```
kant2002, line 4: Are you really sure you want to change only this?

ForNeVeR: Some of the other projects are already migrated. I have no hesitations on upgrading the whole Cesium to net8 (except the runtime library which I still would like to target .NET Framework while possible).

Do you have some other suggestions?

Issue #521, opened 2024-01-22
Preprocessor: BinaryExpression is not correctly evaluate 


7645re:
Test to reproduce the error:
```csharp
    [Fact]
    public Task Test() => DoTest(@"#define NUM 9
#if NUM < 10
int int() {}
#else
int main() {}
#endif
");
```

This line is executed incorrectly - `#if NUM < 10` - `False`

This is because in the EvaluateExpression method, the BinaryExpression class made an error comparing values. It uses a lexicographic comparison that compares strings character by character, so it turns out 9 is greater than 10, only 9 and 1 are compared

```csharp
return Operator switch
        {
            CPreprocessorOperator.Equals => firstValue == secondValue ? "1" : "0",
            CPreprocessorOperator.NotEquals => firstValue != secondValue ? "1" : "0",
            CPreprocessorOperator.LessOrEqual => (firstValue ?? "").CompareTo(secondValue ?? "") <= 0 ? "1" : "0",
            CPreprocessorOperator.GreaterOrEqual => (firstValue ?? "").CompareTo(secondValue ?? "") >= 0 ? "1" : "0",
            CPreprocessorOperator.LessThan => (firstValue ?? "").CompareTo(secondValue ?? "") < 0 ? "1" : "0",
            CPreprocessorOperator.GreaterThan => (firstValue ?? "").CompareTo(secondValue ?? "") > 0 ? "1" : "0",
            CPreprocessorOperator.LogicalAnd => (firstValue.AsBoolean() && secondValue.AsBoolean()) ? "1" : "0",
            CPreprocessorOperator.LogicalOr => (firstValue.AsBoolean() || secondValue.AsBoolean()) ? "1" : "0",
            _ => throw new CompilationException($"Operator {Operator} cannot be used in the preprocessor directives"),
        };
```

it would also be nice to write a handler for the entered numbers, because now it is possible to enter `#define NUM 00`

DOD:
- [ ] Check the other comparison operators
- [ ] Fix calculation errors
- [ ] Write tests for all possible cases 

Optional DOD:
- [ ] Number handler with input verification
- [ ] Handler tests

7645re:
⚔️My chivalrous soul wants to do it myself!

ForNeVeR:
Closed by #531.

PR #531, opened 2024-01-25
Fix BinaryExpression calculation & Tests

7645re:
https://github.com/ForNeVeR/Cesium/issues/521


kant2002:
Can you point where fix to binary expression happens? it's not clear, it seems to be formatting changes. Also maybe not so important,but can you separate formatting chages and functional changes.

7645re:
So far, I have only written tests to understand in which places expression does not work correctly, the logic will be fixed in the following commits

7645re:
suddenly you have any suggestions or suggestions about the tests and their design

Cesium.Preprocessor.Tests/BinaryExpressionTests.cs
```diff
@@ -0,0 +1,72 @@
...
+    [InlineData("0", CPreprocessorOperator.LogicalAnd, "0", false)]
+    [InlineData("0", CPreprocessorOperator.LogicalAnd, "1", false)]
+    [InlineData("1", CPreprocessorOperator.LogicalAnd, "0", false)]
+    [InlineData("1", CPreprocessorOperator.LogicalAnd, "1", true)]
+
+    [InlineData("0", CPreprocessorOperator.LogicalOr, "0", false)]
+    [InlineData("0", CPreprocessorOperator.LogicalOr, "1", true)]
+    [InlineData("1", CPreprocessorOperator.LogicalOr, "0", true)]
+    [InlineData("1", CPreprocessorOperator.LogicalOr, "1", true)]
+    public void EvaluateExpressionAllVariants(
```
ForNeVeR, line 40: _(Minor)_ In general, the test design looks ok to me. Though I would recommend to provide the test with expressions in text and run an actual parser/lexer against them, to ease the maintenance.

Cesium.Preprocessor.Tests/ExpressionTestExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static bool ToBoolean(this string text)
+    {
+        return text switch
+        {
+            "1" => true,
+            "0" => false,
+            _ => throw new NotSupportedException()
+        };
+    }
```
ForNeVeR, line 13: _(Note)_ Just a note that normally `"0"` is considered as falsy, while any other value is considered truthy in C. I have to admit I didn't read the corresponding section on C preprocessor in detail, and maybe we want to do strict checks in our tests.

Cesium.Preprocessor/BinaryExpression.cs
```diff
@@ -2,23 +2,20 @@
 
 namespace Cesium.Preprocessor;
 
-internal sealed class BinaryExpression : IPreprocessorExpression
+public class BinaryExpression(
```
ForNeVeR, line 5: I'd prefer these classes to stay `internal`, but add `InternalsVisibleTo` to this assembly for the tests exclusively.

ForNeVeR (changes requested):

Cesium.Preprocessor.Tests/ExpressionTestExtensions.cs
7645re: Fact, my mistake, sorry :smile:

7645re: Oh, no, I've revised the code, it's just an extension class for ease of testing, this method is not used for calculations in C, but simply translates the result, which is always 0 or 1

ForNeVeR: By the way, see `Cesium.Preprocessor.MacroExtensions.AsBoolean`. Consider reusing that one.

Cesium.Parser.Tests/Cesium.Parser.Tests.csproj
```diff
@@ -45,5 +45,4 @@
         <DependentUpon>PreprocessorTests.cs</DependentUpon>
       </None>
     </ItemGroup>
-
 </Project>
```
7645re, line 48: IDE broke, it turned out by itself, should I return it?

Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
```diff
@@ -550,12 +550,27 @@ public Task IfExpressionCannotConsumeNonInteger()
...
+    [Fact, NoVerify]
+    public async Task IfExpressionLessOrEqualsWhenNotDefinedRight()
+    {
+        var err = await Assert.ThrowsAsync<PreprocessorException>(async () => await DoTest(
+@"#if 1 <= mycondition
+int foo() { return 0; }
+#endif
+"));
+        Assert.Equal("The right-hand element of the expression was not found", err.Message);
+    }
```
7645re, line 573: I remembered that if it did not find a macro in the context, then it will be the true condition, I will correct it now

Cesium.Preprocessor/BinaryExpression.cs
```diff
@@ -2,34 +2,60 @@
...
+                return Convert.ToInt32(macrosValue[2..], 2);
+
+            if (macrosValue.StartsWith("0x"))
+                return Convert.ToInt32(macrosValue[2..], 16);
+
+            if (macrosValue[0] == '0' && char.IsDigit(macrosValue[1]))
+                return Convert.ToInt32(macrosValue[2..], 8);
+
+            throw new NotSupportedException("Invalid macros format");
+        }
```
7645re, line 59: I want to redo it in regex

Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
```diff
@@ -732,4 +739,49 @@ public Task IfExpressionCannotConsumeNonInteger()
...
+        string oper,
+        string secondValue,
+        bool expectedResult)
+    {
+        const string funcText = "int foo() { return 0; }";
+        var actualResult = await DoPreprocess($@"#if {firstValue} {oper} {secondValue}{Environment.NewLine}"
+                                        + funcText
+                                        + $"{Environment.NewLine}#endif");
+        Assert.Equal(expectedResult ? funcText : "", actualResult.Trim());
+    }
```
7645re, line 786: redo using [MemberData]

ForNeVeR (approved):
Thanks! This is an improvement over the previous version, though it caused a regression (the one for that we've had no tests beforehand).

I am going to merge this and open a follow-up issue.

ForNeVeR:
The newly discovered problem turns out to be a pretty minor one, and also not a regression at all: #534

PR #533, opened 2024-01-27
Update TokenExtensions.cs

Evgeny608799:
Modification of `current >= '0' && current <= '9'` to `Char.IsDigit()`

### DIFF
--- a/Cesium.Parser/TokenExtensions.cs
+++ b/Cesium.Parser/TokenExtensions.cs
@@ -89,14 +89,14 @@ public static class TokenExtensions
                             char current = result.ElementAt(i + counter);
                             do
                             {
-                                int digit = current >= '0' && current <= '9' ? current - '0' : (char.ToUpperInvariant(current) - 'A') + 10;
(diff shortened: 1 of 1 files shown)


ForNeVeR:
Sadly, I don't believe this change is correct, I am sorry :(

`Char.IsDigit` is documented to accept Unicode digits as well (for example, `६` is a digit as well!), while the C standard only mandates support for the standard digits `0…9`.

I would propose to update it and extract the same check for `0…9` to a common method.

ForNeVeR (changes requested):
Sadly, I don't believe this change is correct, I am sorry :(

`Char.IsDigit` is documented to accept Unicode digits as well (for example, `६` is a digit as well!), while the C standard only mandates support for the standard digits `0…9`.

I would propose to update it and extract the same check for `0…9` to a common method.

Evgeny608799:
Replaced `Char.IsDigit` with `Char.IsAsciiDigit` to check digits only in **ASCII**

ForNeVeR (approved):
Thanks!

PR #544, opened 2024-02-08
Add _Bool type

kant2002:


ForNeVeR:
[We have it documented](https://github.com/ForNeVeR/Cesium/blob/main/docs/type-system.md) to correspond to `System.Bool`, though. Are you sure we want to change that decision?

ForNeVeR:
Tests say
```
Right operand of type PrimitiveType { Kind = Bool } is not numeric.
```

Heh. Need help with that?

kant2002:
> Heh. Need help with that?

Compasion with my humane limitations would be enough!

ForNeVeR (approved):
Thanks!

Issue #225, opened 2022-08-14
Object initializers


ForNeVeR:
Seemingly, we don't support this:

```c
typedef struct { int x; } foo;
foo xx = { 1 };
```

We should support it.

Look for the number `xxx` in the code to find clues to implement this feature.

PR #549, opened 2024-02-26
add object initialization

BadRyuner:
Partially covers #225
I just don't know how to implement initialization of array fields. (I marked this with NoVerify). It requires some tricks with the parser.

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
+++ b/Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
@@ -41,7 +41,7 @@ internal sealed class CompoundObjectInitializationExpression : IExpression
         var fieldsDefs = typeDef.Fields;
         var initializers = _initializers;
 
-        if (typeDef.Name.Contains("<SyntheticBuffer>"))
+        if (typeDef.IsCArray())
         {
             var element = typeDef.Fields[0].FieldType;
             for (int i = 0; i < initializers.Length; i++)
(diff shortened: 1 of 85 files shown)


BadRyuner:
Oh, I forgot about ParserTests. Integration Test with complex structure initialization works like all other IntegrationTests and CodeGenTests.

ForNeVeR:
Alright, please take a look at tests and signal if you need help.

BadRyuner:
The main reason for the failed tests is that I added Designation to the Initializer. As the test checks the new AST with the old AST, it triggers on the fact that "Designation: null" appears. Should I just accept these new tests through the script?

ForNeVeR:
> Should I just accept these new tests through the script?

Yep!

Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
```diff
@@ -0,0 +1,260 @@
...
+    {
+        if (_type == null && _typeDef == null)
+            throw new Exception("_type is null!");
+
+        var instructions = scope.Method.Body.Instructions;
+        TypeDefinition typeDef = _type != null ? ((StructType)_type).Resolve(scope.Context).Resolve() : _typeDef!;
+        var fieldsDefs = typeDef.Fields;
+        var initializers = _initializers;
+
+        if (typeDef.Name.Contains("<SyntheticBuffer>"))
```
ForNeVeR, line 44: I'd suggest to introduce some helper method that would do this check for us. I am not comfortable with the idea of random pieces of code just performing type classification based on type names.

Let's add it to `TypeSystemEx` perhaps?

ForNeVeR (commented):
I am still in the process of review, but decided to leave some comments already.

Cesium.CodeGen/Ir/Types/StructType.cs
```diff
@@ -90,6 +92,14 @@ public void FinishEmit(TypeDefinition definition, string name, TranslationUnitCo
...
             definition.Fields.Add(field);
         }
+
+        // emit ctor
+        Constructor = new MethodDefinition(".ctor", MethodAttributes.Public | MethodAttributes.HideBySig
+            | MethodAttributes.RTSpecialName | MethodAttributes.SpecialName, context.TypeSystem.Void);
+        var body = new MethodBody(Constructor);
+        Constructor.Body = body;
+        body.Instructions.Add(Instruction.Create(OpCodes.Ret));
+        definition.Methods.Add(Constructor);
```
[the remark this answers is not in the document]
BadRyuner: I can replace newobj with initobj, this would then eliminate the need for a constructor.

Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
BadRyuner: Like: `TypeSystemEx.IsArray(TypeRefernce typeRef) => typeRef.Name.StartsWith("<SyntheticBuffer>");` ?

Cesium.CodeGen/Ir/Types/StructType.cs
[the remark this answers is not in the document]
ForNeVeR: Well yeah, great, let's do it!

Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
ForNeVeR: Yep, exactly like this.

ForNeVeR (approved):
Thanks!

Issue #511, opened 2024-01-05
`#pragma pinvoke`


ForNeVeR:
## The Problem
Real-world Cesium users will inevitably encounter a range of problems involving interop with native code.

1. They might want to convert their existing programs to Cesium file by file, while keeping interoperability between the managed and unmanaged parts (`#pragma unmanaged`, looking at you right now!).

   This will be useful in cases when some language features are not yet supported in Cesium, or if certain translation units are too reliant on native dependencies.

   Cesium may support that by compiling the managed part of the code and providing an ability to dynamically link to its native part. In this case, the managed and native parts of the program would share the same set of the header files, but the classical _linking_ stage would be augmented by Cesium stepping in and wiring it all via P/Invoke.
3. They might want to use some native dependencies for their programs. _Of course_ they will.
 
   The classical solution from the native world is for the libraries to provide native link modules (for example, `.lib` on Windows). The compiler will just note that certain functions are not known at the compilation stage, and will delay the resolution to the linker. And the linker would analyze the set of available link modules and seek for the mentioned dependencies among them.

   Cesium is unable to use that solution because we cannot link managed and native parts together. But CLI has an ability to P/Invoke. Let's use that!

Currently, it is theoretically possible to interop with this code using Cesium, but it is very clunky and far from being a good solution. In particular, you may take the following strategy:

1. Build a native library somehow.
2. Write a C# wrapper around it using P/Invoke.
3. Write a Cesium wrapper around C# wrapper using `__cli_import`, e.g. `__cli_import("MyClass::MyFunction") int MyFunction(int x)` — and that for every function.
4. Build a Cesium program while referencing the C# library that P/Invokes the native library.

While possible in theory, this is, as you can imagine, not very practical.

## The Solution
- [x] Introduce a new `#pragma pinvoke` that is invoked in two forms.
   1. `#pragma pinvoke("mydll.dll")`. This command would make all the following function declarations (if not provided by the current translation unit) to be looked up in `mydll.dll`, effectively turning them into P/Invoke declarations. For example, this program:
      ```c
      #pragma pinvoke("mydll.dll")
      int foo_bar(int*);
      ```
      would be an equivalent of the following C# code:
      ```csharp
      // partial static class MyCurrentTranslationUnit
      [DllImport("mydll.dll")]
      static extern int foo_bar(int* _a);
      ```
   6. `#pragma pinvoke(end)` that will restore the default resolve behavior.

[...]

PR #550, opened 2024-02-28
Add #pragma pinvoke

BadRyuner:
Yes, this is not a good implementation of "#pragma pinvoke"... (#511)
Some methods I didn't name correctly, hope for some help with that!

But it's working just fine.


BadRyuner:
Need a good example of a pinvoke function call that is on all three platforms in the same library... On Windows "puts" function in mscvrt, on Linux "puts" function in libc...

BadRyuner:
Well, the tests passed.

Cesium.CodeGen/Contexts/GlobalConstructorScope.cs
```diff
@@ -77,4 +80,30 @@ public Instruction ResolveLabel(string label)
...
+
+    public T? GetSpecialEffect<T>()
+    {
+        for(int i = _specialEffects.Count - 1; i >= 0; i--)
+        {
+            var effect = _specialEffects[i];
+            if (effect is T t)
+                return t;
+        }
+        return default;
```
ForNeVeR, line 94: Would this work?
```suggestion
        return _specialEffects.OfType<T>().FirstOrDefault();
```

Cesium.CodeGen/Contexts/IDeclarationScope.cs
```diff
@@ -45,6 +46,10 @@ internal interface IDeclarationScope
     /// </summary>
     /// <returns>Name of the virtual label which can be used by continue statement</returns>
     string? GetContinueLabel();
+    
+    void PushSpecialEffect(object declaration);
```
ForNeVeR, line 50: Please add a comment indicating our pragma as an example of this "special effect", for future readers.

Cesium.CodeGen/Contexts/Meta/FunctionInfo.cs
```diff
@@ -19,6 +20,8 @@ internal record FunctionInfo(
     public StorageClass StorageClass { get; set; } = StorageClass;
     public bool IsDefined { get; set; } = IsDefined;
     public string? CliImportMember { get; set; }
+    public string? DllLibraryName { get; set; }
+    public string? DllImportNameStrip { get; set; }
```
ForNeVeR, line 24: Let's add documentation referencing `PInvokeDefinition.Prefix`, for the reader to get why are there two properties.

Cesium.Preprocessor/CPreprocessor.cs
```diff
@@ -284,6 +298,12 @@ await foreach (var item in subProcessor.GetPreprocessingResults())
         yield return new Token<CPreprocessorTokenType>(new Range(), new Location(), "\n", NewLine);
     }
 
+    private static IEnumerable<IToken<CPreprocessorTokenType>> TokenizeString(string code)
+    {
+        var tokenizer = new CPreprocessorLexer("c:/null.c", code);
```
ForNeVeR, line 303: Let's either mark this as a `null` path (making the parameter nullable), or use some more specific name, like `<null>` or something.

ForNeVeR (changes requested):
Great work! I've left some comments and ready to discuss minor issues.

Cesium.CodeGen.Tests/CodeGenPinvokeTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            var preprocessor = new CPreprocessor(
+                _mainMockedFilePath,
+                lexer,
+                includeContext,
+                definesContext,
+                warningProcessor);
+            var result = await preprocessor.ProcessSource();
+            return result;
+        }
+    }
```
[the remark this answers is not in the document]
BadRyuner: This is from another test: [Cesium.Parsers.Tests->PreprocessorTests](https://github.com/ForNeVeR/Cesium/blob/63bd762e43f52dfd6cfb4ed1478fe190f84cd1fc/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs#L26)
They don't overlap

ForNeVeR (approved):
Thanks!

PR #553, opened 2024-03-06
FunctionInfo.cs to immutable

Evgeny608799:
FunctionInfo to immutable and refactoring code to maintain immutable behavior. **Issue #489**

### DIFF
--- a/Cesium.CodeGen/Contexts/AssemblyContext.cs
+++ b/Cesium.CodeGen/Contexts/AssemblyContext.cs
@@ -12,6 +12,7 @@ using Cesium.Core;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Cecil.Rocks;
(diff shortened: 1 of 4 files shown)


kant2002:
I think we want to change a way how compiler collect information about function info. That’s essentially the same as before IMO

ForNeVeR (changes requested):
Well, yeah. Unfortunately, a deeper redesign was required in this task. It is not as simple as "replace mutable function data with builder, and replace the data in dictionary instead of mutation".

Several problems with this PR:
- tests are not passing
- you introduced `ICloneable`, while C# records are perfectly (and type-safely!) cloneable already, and this type was already a record, and its cloneability was already utilized in certain places — but this PR just outright replaces that

I am very sorry to say this, but I don't think the current approach will lead to addressing the initial issue.

PR #558, opened 2024-03-22
Improve Main function signature equivalency

BadRyuner:
Now main functions using const int argc or const char * cosnt * argv or char * cont * argv are valid.
Issue #66


ForNeVeR (approved):
Thanks!

PR #559, opened 2024-03-23
Add unary plus

BadRyuner:
Also, unary plus is now supported. I've added two tests with it.

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
@@ -111,6 +111,16 @@ int main() { foo x,x2; x2.x=0; }");
 ");
 
     [Fact]
+    public Task UnaryPlusAndStart() => DoTest(@"
+int main() {
+    short a = -2;
+    short* b = &a;
+    short c = *b;
+    int x = (+c) - (+1) - (-1);
+    return sizeof(+c);
+} ");
+
+    [Fact]
     public Task ReturnWithoutArgument() => DoTest(@"void console_read()
 {
     return;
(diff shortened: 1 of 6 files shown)


Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -332,7 +333,7 @@ public static int FPrintF(void* stream, byte* str, void* varargs)
                 case "f":
                     {
                         var floatNumber = ((double*)varargs)[consumedArgs];
-                        string floatNumberString = floatNumber.ToString("F" + (trim == -1 ? 6 : trim));
+                        string floatNumberString = floatNumber.ToString("F" + (trim == -1 ? 6 : trim), CultureInfo.InvariantCulture);
```
ForNeVeR, line 336: Why have you added this here? Does it mess up with tests or something?

Anyway, I do not believe this is correct in a general sense.

This particular module is written to respect the current culture, keeping in mind that we'll integrate the C locale system with .NET's cultures eventually.

I would actually prefer to have invariant culture by default, but
- that's a discussion I'd like to separate from implementation if the unary plus operator
- if anything, then Cesium would have to set the current culture in `main()`

ForNeVeR (changes requested):

Cesium.Runtime/StdIoFunctions.cs
BadRyuner: Because without adding CultureInfo.InvariantCulture, Cesium.PrintF fails tests on non-InvariantCulture systems, which doesn't match the output in c.
image

ForNeVeR: Ok, so all this business is incredibly wacky. I would still ask you to roll this change back, and let's discuss and reintroduce some proper mode in #560. This change should not indeed be related to the unary plus operator, and also to my conclusion in #560 it is also not the correct one.

BadRyuner: Done

PR #562, opened 2024-04-20
Improved cli-interop

BadRyuner:
A FuncPtr can now be turned into a corresponding Delegate and also a Delegate can be turned into a FuncPtr.
You can also use UTF8String instead of byte*. This is more convenient and it contains the fastest implementations of some methods for the corresponding .Net versions.

### DIFF
--- a/Cesium.Runtime/FuncPtr.cs
+++ b/Cesium.Runtime/FuncPtr.cs
@@ -12,14 +12,9 @@ public readonly unsafe struct FuncPtr<TDelegate> where TDelegate : MulticastDele
         _value = (IntPtr)ptr;
     }
 
-    public FuncPtr(IntPtr ptr)
-    {
-        _value = ptr;
-    }
-
     public static implicit operator TDelegate(FuncPtr<TDelegate> funcPtr) => (TDelegate)Activator.CreateInstance(typeof(TDelegate), [null, funcPtr._value])!;
     public static implicit operator FuncPtr<TDelegate>(TDelegate @delegate) => @delegate.Method.MethodHandle.GetFunctionPointer();
-    public static implicit operator FuncPtr<TDelegate>(IntPtr funcPtr) => new(funcPtr);
+    public static implicit operator FuncPtr<TDelegate>(IntPtr funcPtr) => new((void*)funcPtr);
     public static implicit operator FuncPtr<TDelegate>(void* funcPtr) => new(funcPtr);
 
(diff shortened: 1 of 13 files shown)


Cesium.Runtime/FuncPtr.cs
```diff
@@ -1,14 +1,23 @@
+using System.Runtime.InteropServices;
+
 namespace Cesium.Runtime;
 
 /// <summary>A class encapsulating a C function pointer.</summary>
-public readonly unsafe struct FuncPtr<TDelegate> where TDelegate : Delegate // TODO[#487]: Think about vararg and empty parameter list encoding.
+public readonly unsafe struct FuncPtr<TDelegate> where TDelegate : MulticastDelegate // TODO[#487]: Think about vararg and empty parameter list encoding.
 {
-    private readonly long _value;
+    private readonly IntPtr _value;
```
ForNeVeR, line 8: Please revert back to `long`. This will break portability to x86 systems.

Cesium.Runtime/UTF8String.cs
```diff
@@ -0,0 +1,139 @@
...
+/// <summary>
+/// A useful wrapper over UTF8 strings.
+/// </summary>
+[EquivalentType(typeof(byte*))]
+[StructLayout(LayoutKind.Sequential)]
+public unsafe readonly struct UTF8String
+{
+    public readonly static UTF8String NullString = new UTF8String((byte*)0);
+
+    private readonly byte* _value;
```
ForNeVeR, line 17: If we want this to be used in portable code, then let's change it to `long`.

Cesium.Runtime/VoidPtr.cs
```diff
@@ -3,14 +3,14 @@ namespace Cesium.Runtime;
 /// <summary>A class encapsulating an opaque pointer (aka <code>void*</code> in C).</summary>
 public readonly unsafe struct VoidPtr
 {
-    private readonly long _value;
+    private readonly IntPtr _value;
```
ForNeVeR, line 6: No, this is not correct. `VoidPtr` is supposed to have a size of `long` and not `IntPtr`: 8 bytes on every system. It is used for portable code, for the "wide arch".

For any other arch, we just use normal pointers.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #575, opened 2024-04-22
[sdk]: Possible fix missing/outdated artifacts before compiler packing

seclerp:
Not sure about the root cause, but this may probably fix some weird issues when running compiler targets subsequently


ForNeVeR (approved):
Thanks!

PR #576, opened 2024-04-22
Add ##__VA_ARGS__ and __VA_OPT__

BadRyuner:
Examples for the tests I took from here
https://en.cppreference.com/w/c/preprocessor/replace#Function-like_macros
The code passed all the tests


Cesium.Preprocessor/MacroExpansionEngine.cs
```diff
@@ -268,6 +272,18 @@ IEnumerable<IToken<CPreprocessorTokenType>> ClearSpaceBuffer()
         }
         yield break;
 
+        void PoofComma()
+        {
+            foreach (var tok in lexer.EnumerateInReverse()) // look at previous tokens
+                if (tok.Text == ",") // is that a comma?
+                {
+                    MemoryMarshal.GetReference(tok.Text.AsSpan()) = ' '; // poof and there`s no more comma
```
ForNeVeR, line 280: This modifies a `string` in-place which may have consequences. Please avoid doing that. All the compiler structures, including AST, lexer data and whatnot are immutable, not even starting about `string`.

Cesium.Preprocessor/MacroExpansionEngine.cs
```diff
@@ -10,6 +11,8 @@ namespace Cesium.Preprocessor;
 
 public class MacroExpansionEngine(IWarningProcessor warningProcessor, IMacroContext macroContext)
 {
+    private string __VA_OPT__Rule = ",";
```
ForNeVeR, line 14: I am not quite sure about this state introduced into the engine. Note that it's reused between different preprocessor calls; it wasn't meant to have any state. And here you introduce a variable that may change while expanding `__VA_OPT__`. What if we need to expand it twice? What if we need to expand it in different macros?

I'd suggest avoid relying on this kind of state.

ForNeVeR (changes requested):

Cesium.Preprocessor/MacroExpansionEngine.cs
BadRyuner: I don't see any other way to remove the previous token if we create a new collection via yield return that we don't have access to.

ForNeVeR: Alright, I'll add this to my queue to rearchitecture the corresponding part.

Cesium.Preprocessor/MacroExpansionEngine.cs
BadRyuner: Then add VA_OPT as a field in MacroArguments?

PR #588, opened 2024-05-18
[templates]: Add basic project templates for Cesium

seclerp:
I've added 2 templates for Cesium:
- `cesiumapp`: like `consoleapp` but with Cesium
- `cesiumlib`: like `classlib` but with Cesium

Both templates have an option to configure a target framework for it

I've also added a Nuke target to pack a NuGet package for a template project, so it's ready to be used once the CI/CD infrastructure for SDK publishing is ready


ForNeVeR (approved):
Thanks!

PR #597, opened 2024-06-01
[sdk]: Include `.c` and `.h` items by default

### LINKED ISSUE #574: SDK: include all the `.c` files by default?
Currently, we rely on user adding the `.c` files into the project themselves.

Should we just go ahead and include `**/*.c` as `Compile` and `**/*.h` as `None` (to be visible in the IDE) by default?

Are there any concerns about this? I'd say we should do it.

seclerp:
Fixes #574

Instead of just ignoring all default items with `EnableDefaultItems=false` I decided to switch off all particular default item groups except the ones included in `Compile` items. This allows us to be better integrated with common SDK properties.

Since `DefaultLanguageSourceExtension` only supports 1 extension, others like `.h` were added separately with the same switching logic.

For the reference I used [Microsoft.NET.Sdk.DefaultItems.props](https://github.com/dotnet/dotnet/blob/release/8.0.1xx/src/sdk/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props)


ForNeVeR:
CI seems unhappy with the changes. Could you please take a look?

ForNeVeR:
There was a problem likely caused by us caching the templates produced during tests, in system-wide nuget cache. I'll take a look at that: we should not use the system cache during template/sdk tests.

ForNeVeR (approved):
Thanks!

PR #617, opened 2024-07-26
Infinite loop check for return

kant2002:
Current Cesium think that this loop is never finished.

```
int test() {
  for(;;) {
     return 5;
  }
}
```


kant2002:
I still have to deal with scope merge. Dumb strategy does not work, and I have no clear idea what should I do now.

Cesium.CodeGen.Tests/CodeGenMethodTests.cs
```diff
@@ -348,6 +348,9 @@ public void ImplicitReturnDisallowedNonMain() => DoesNotCompile(@"int foo()
     int unused;
 }", "Function foo has no return statement.");
 
+    [Fact]
+    public Task ReturnInsideInfiniteForTest() => DoTest("int test() { for(;;) { return; } }");
```
ForNeVeR, line 352: _(Note)_ Codegen for this method is quite suspicious:
```
System.Int32 <Module>::test()
  IL_0000: nop
  IL_0001: ret
  IL_0002: nop
  IL_0003: br IL_0000
  IL_0008: nop
```
`ret` seems to be the first executed instruction.

Can confirm it was always like this (even before your PR).

I'm making a note to open an issue about that after merge.

Cesium.CodeGen/Contexts/VariableInfo.cs
```diff
@@ -6,4 +6,6 @@ namespace Cesium.CodeGen.Contexts;
 
 internal record VariableInfo(string Identifier, StorageClass StorageClass, IType Type, IExpression? Constant)
 {
+    static int CurrentIndex = 0;
```
ForNeVeR, line 9: _(Note)_

I'd like to have some sort of indexing that's tied to the translation unit, or something.

Such a counter is too error-prone: it will easily overflow on 2 billion variables constructed in a single compilation session. Moreover, this construct makes it impossible to copy a `VariableInfo`.

I am not against having this for now, but we'll need to get rid of it eventually.

ForNeVeR (approved):
Thanks!

Cesium.CodeGen.Tests/CodeGenMethodTests.cs
ForNeVeR: #633.

Cesium.CodeGen/Contexts/VariableInfo.cs
ForNeVeR: #634.

PR #632, opened 2024-08-31
Fix issue with empty initializer

kant2002:


ForNeVeR:
For `structs/struct_init.c`, this yields a runtime error:
```
[ERR] Unhandled exception. System.InvalidProgramException: The JIT compiler encountered invalid IL code or an internal limitation.
[ERR]    at main()
[ERR]    at <SyntheticEntrypoint>()
[ERR] 
Command exit code: -532462766
```

ForNeVeR:
I think @kant2002 mentioned he still works on this one.

Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
```diff
@@ -51,26 +55,42 @@ public void EmitTo(IEmitScope scope)
...
+                {
+                    _prefixAction();
+                    instructions.Add(Instruction.Create(OpCodes.Ldflda, _typeDef));
+                }
                 if (i != 0)
                 {
                     instructions.Add(Instruction.Create(OpCodes.Ldc_I4, i));
                     instructions.Add(Instruction.Create(OpCodes.Sizeof, element)); // size = sizeof(array element)
                     instructions.Add(Instruction.Create(OpCodes.Mul)); // offset = id * size
+                    instructions.Add(Instruction.Create(OpCodes.Add)); // offset = id * size
```
ForNeVeR, line 69: I feel like the comment is wrong. What does this actually do?

Cesium.CodeGen/Ir/Expressions/CompoundObjectInitializationExpression.cs
```diff
@@ -51,26 +55,42 @@ public void EmitTo(IEmitScope scope)
...
 
         instructions.Add(Instruction.Create(OpCodes.Ldloca, newobj));
         instructions.Add(Instruction.Create(OpCodes.Initobj, typeDef));
 
         if (initializers.Length == 0) // zero init like SomeType name = { };
+        {
+            instructions.Add(Instruction.Create(OpCodes.Ldloc, newobj)); // push new object
+            if (_prefixAction is not null)
+            {
+                _prefixAction();
```
ForNeVeR, line 89: Is it okay that `_prefixAction` gets called twice? Let's maybe write some comment about its purpose and possible behavior?

ForNeVeR (changes requested):

ForNeVeR (approved):
I have to say that I don't like the business around `prefixAction` and `postfixAction` boo much, but if it works, then it works.

Thanks!

PR #635, opened 2024-09-01
Build: enable package locks

ForNeVeR:
Not saying this is too practical, but let's give it a try.


ForNeVeR:
Ok, not gonna happen. NuGet package locking is, sadly, still not working in .NET ecosystem.

PR #637, opened 2024-09-05
Improve code generation for functions

### LINKED ISSUE #633: Endless loop compiled with ret in the beginning
_Originally posted by @ForNeVeR in https://github.com/ForNeVeR/Cesium/pull/617#discussion_r1740187552_

https://github.com/ForNeVeR/Cesium/blob/f1b5204e9d1f0869ed8c4ee0caf3c3e41ad6a5a9/Cesium.CodeGen.Tests/CodeGenMethodTests.cs#L351-L352

Codegen for this method is quite suspicious:
```
System.Int32 <Module>::test()
  IL_0000: nop
  IL_0001: ret
  IL_0002: nop
  IL_0003: br IL_0000
  IL_0008: nop
```
`ret` seems to be the first executed instruction.

This shouldn't be so.

kant2002:
- Allow missing return for non void functions.
- Make sure that unreachable code at the end of file has ret instruction.

Closes #633

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenMethodTests.cs
@@ -342,14 +342,23 @@ int main()
     int unused = 0;
 }");
 
-    [Fact, NoVerify]
-    public void ImplicitReturnDisallowedNonMain() => DoesNotCompile(@"int foo()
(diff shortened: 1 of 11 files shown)


ForNeVeR (approved):
Thanks!

PR #643, opened 2024-09-06
Fix bug with codegeneration with forward declaration

kant2002:
when type was forward declared before child type which used inside that declared type later on, codegen was broken. This lead to simplification of the struct emit, since no need to special case anon types. This lead to changes in naming which I think is reasonable, since that way always anon types prefixed with <typedef>


ForNeVeR (approved):
Thanks!

PR #644, opened 2024-09-06
Add support for implicit marshalling of the eq types

kant2002:
Also MSVC does not have support for strndup which is C23


Cesium.Runtime/StringFunctions.cs
```diff
@@ -192,4 +193,19 @@ public static int MemCmp(void* lhs, void* rhs, nuint count)
...
+    }
+
+    public static UTF8String StrNDup(UTF8String src, nuint count)
+    {
+        if (src.Pointer == null)
+            return UTF8String.NullString;
+
+        var dest = new UTF8String((byte*)StdLibFunctions.Malloc(count));
+        src.CopyTo(dest, count);
+        return dest;
```
ForNeVeR, line 209: I think this misses the null terminator. Please note this wording from the standard:
> If the array pointed to by `s` does not contain a null within the first size
characters, a null is appended to the copy of the array.

Meaning the result should always be null-terminated, regardless of `src[count]` being a null byte or not.

ForNeVeR (changes requested):

ForNeVeR:
I have to say: I am quite confused about all this `RESOLUTION_CESIUM` business, but let's make this a topic of a new issue, after we finish with this PR.

kant2002:
Re: `RESOLUTION_CESIUM`. I almost manage to make it work, except it fails during runtime, since Cesium.Runtime cannot be resolved. I did not pursue investigation, current way of things was present originally. Need time to investigate and fix properly, maybe something simple.

ForNeVeR (approved):

ForNeVeR:
Thanks!

PR #645, opened 2024-09-06
Allow initialization of char ptr array

kant2002:


Cesium.CodeGen.Tests/CodeGenArrayTests.cs
```diff
@@ -197,4 +197,21 @@ int main(void)
...
+    public Task ConstByteArrayTest() => DoTest(@"
+int main() {
+    const char a[1] = { 255 };
+}");
+
+    [Fact]
+    public Task ArrayCharComparison() => DoTest(@"
+int main() {
+    const char a[1] = { 'A' };
+    if (a[0] != 'D') {
```
slyshykO, line 211: Why is there a check for 'D' despite the array being initialized with 'A'?

kant2002: I test what’s generated. I need comparison with any char literal. Does not matter what.

ForNeVeR: Yeah. This doesn't matter. We don't even run this code. It's for the purpose of dumping the codegen results. So, more variety (within some measure of course) — more chances the tests will catch unexpected codegen changes. This is good.

ForNeVeR (approved):
Thanks!

PR #659, opened 2024-09-20
Allow const eval expression in array declarations

kant2002:


Cesium.CodeGen/ConstantEvaluator.cs
```diff
@@ -2,71 +2,84 @@
...
-                        UnaryOperator.Negation => new IntegerConstant(-constInt.Value),
-                        UnaryOperator.BitwiseNot => new IntegerConstant(~constInt.Value),
-                        UnaryOperator.LogicalNot => new IntegerConstant(constInt.Value != 0 ? 0 : 1),
-                        UnaryOperator.AddressOf or UnaryOperator.Indirection => throw new CompilationException($"Operator {unOp.Operator} is not compile-time evaluable"),
-                        _ => throw new ArgumentOutOfRangeException($"Invalid unary operator {unOp.Operator}"),
+                        UnaryOperator.Negation => (null, new IntegerConstant(-constInt.Value)),
+                        UnaryOperator.BitwiseNot => (null, new IntegerConstant(~constInt.Value)),
+                        UnaryOperator.LogicalNot => (null, new IntegerConstant(constInt.Value != 0 ? 0 : 1)),
+                        UnaryOperator.AddressOf or UnaryOperator.Indirection => ($"Operator {unOp.Operator} is not compile-time evaluable", null),
+                        _ => ($"Invalid unary operator {unOp.Operator}", null),
```
ForNeVeR, line 43: I suggest keeping this one as an `ArgumentOutOfRangeException`.

While the others are, indeed, compilation errors — something that the user should be able to correct, this one in compiler implementation issue, and should not be mixed with the others. It should be thrown immediately, on sight.

ForNeVeR (changes requested):

ForNeVeR (approved):
Thanks!

PR #678, opened 2024-10-29
Add option to dump AST

kant2002:
I choose not JSON, to have more dense format, since JSON too sparse to navigate.


ForNeVeR:
To my taste, the solution is too ad-hoc, but for now I approve. Will open a separate issue on automating of this. Perhaps we can get auto-generation from JSON or something like that.

Also, maybe a visitor over all the AST would help, at the end :)

ForNeVeR (approved):

PR #679, opened 2024-10-29
Implement fopen, fgets, fclose, rewind, fseet, feof

kant2002:


ForNeVeR:
Follow-up: #681.

ForNeVeR (approved):
Thanks!

PR #685, opened 2024-10-30
Add missing escape symbols

kant2002:
Also ignore \ before unknown escape char.

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/Constants/CharConstant.cs
+++ b/Cesium.CodeGen/Ir/Expressions/Constants/CharConstant.cs
@@ -1,6 +1,6 @@
+using System.Globalization;
 using Cesium.CodeGen.Contexts;
 using Cesium.CodeGen.Ir.Types;
-using Cesium.Core;
 using Mono.Cecil.Cil;
 
 namespace Cesium.CodeGen.Ir.Expressions.Constants;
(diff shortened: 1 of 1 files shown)


Cesium.CodeGen/Ir/Expressions/Constants/CharConstant.cs
```diff
@@ -41,10 +41,11 @@ private static char UnescapeCharacter(string text)
             'n' => '\n',
             'r' => '\r',
             't' => '\t',
-            'v' => '\v',
+            '?' => '?',
+            'v' => (char)0x0b,
```
ForNeVeR, line 45: Huh, is there a reason you've decided to replace `'\v'` with `0x0b`? As far as I see, in C# it's the same thing, so it should work as it was 😅

Cesium.CodeGen/Ir/Expressions/Constants/CharConstant.cs
```diff
@@ -41,10 +41,11 @@ private static char UnescapeCharacter(string text)
...
             'x' => (char)int.Parse(text.AsSpan(2), System.Globalization.NumberStyles.AllowHexSpecifier),
             > '0' and < '9' => (char)Convert.ToInt32(text.Substring(2), 8),
-            _ => throw new CompilationException($"Unknown escape sequence '{text}'"),
+            _ => text[1],
```
ForNeVeR, line 48: Let's warn on it at least? This is not valid according to the standard, even though some compilers indeed get it passed with a warning.

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Expressions/Constants/CharConstant.cs
ForNeVeR: Ok I dig into this for a bit, and decided to open a new issue on it. Anyway, this deserves some more polishing, let's do it separately.

ForNeVeR: See #699.

ForNeVeR (approved):
Thanks!

PR #691, opened 2024-11-03
Distinguish identifiers and indirection which detected as typecasts

kant2002:
Currently compiler do not know, if `(x)` is typecase or some identifier in the brackets, so we probe registered types. Also ignore constant specifiers when trying to find constant property on the type.

### DIFF
--- a/Cesium.CodeGen/Ir/Expressions/BinaryOperators/BinaryOperatorExpression.cs
+++ b/Cesium.CodeGen/Ir/Expressions/BinaryOperators/BinaryOperatorExpression.cs
@@ -60,7 +60,9 @@ internal sealed class BinaryOperatorExpression : IExpression
         }
 
         if (!Operator.IsComparison() || (leftType.IsNumeric() && rightType.IsNumeric() && !leftType.IsEnum() && !rightType.IsEnum() && !leftType.IsBool() && !rightType.IsBool()))
-        { 
+        {
+            leftType = leftType.EraseConstType();
+            rightType = rightType.EraseConstType();
             var commonType = TypeSystemEx.GetCommonNumericType(leftType, rightType);
             if (!leftType.IsEqualTo(commonType))
             {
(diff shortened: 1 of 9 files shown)


Cesium.CodeGen/Ir/Expressions/BinaryOperators/BinaryOperatorExpression.cs
```diff
@@ -173,6 +175,8 @@ public IType GetExpressionType(IDeclarationScope scope)
 
         // both bitwise and arithmetic operators obey same arithmetic conversions
         // https://en.cppreference.com/w/c/language/operator_arithmetic
+        leftType = leftType.EraseConstType();
+        rightType = rightType.EraseConstType();
```
ForNeVeR, line 179: Hey, what do you think of moving this inside of `TypeSystem.GetCommonNumericType`?

Otherwise, every client of that class will have to call these to handle `const`.

In particular, currently we have `ConditionalExpression` calling it as well — I bet we'll have problems comparing const and non-const numbers then.

ForNeVeR (changes requested):

Cesium.CodeGen/Ir/Expressions/BinaryOperators/BinaryOperatorExpression.cs
kant2002: Not sure about it. I think `ConstType` as type is bad decision. My current thinking is that `IsConst` shoud be property of all type.

Also it's only 2 places. I want to avoid generalization when possible. I also do not have mental model why const correctness important for casting everywhere, and that logic may, or may not, be related to place where we call this function

ForNeVeR: Alright. I thought about this a bit, and I am not against making constness a property of a type. But in either case, we'll need a set of helper methods to compare/check assignability/get rid of `const`, so I'd not believe it would bring us much value either.

ForNeVeR: Regarding the problem itself — let's leave this for another time when somebody will want to compare a const and a non-const number :)

PR #703, opened 2024-11-10
Add parsing compound literal expressions

kant2002:
Also improve parser to closer match C spec

Related #207


ForNeVeR (approved):
Thanks!

Issue #211, opened 2022-08-14
Parse all the C declarations


ForNeVeR:
We should fully implement the following section of the C standard:

- [ ] 6.7 Declarations
  - [ ] 6.7.1 Storage-class specifiers
  - [ ] 6.7.2 Type specifiers
  - [x] 6.7.3 Type qualifiers
  - [ ] 6.7.4 Function specifiers
  - [ ] 6.7.5 Alignment specifier
  - [ ] 6.7.6 Declarators
  - [ ] 6.7.7 Type names
  - [x] 6.7.8 Type definitions
  - [ ] 6.7.9 Initialization
  - [ ] 6.7.10 Static assertions

PR #705, opened 2024-11-11
Add parsing of function specifiers

### LINKED ISSUE #706: 6.7.4 Function specifiers

kant2002:
They are harless in a sense that we can not even implement them. Inline can be taken care by JIT, and _Noreturn would be handled later, since it requrie control flow analysys which we are barely have and don't have expertise at all.

Related #211
Closes #706


ForNeVeR (commented):
Yep, great!

PR #708, opened 2024-11-11
Support inline and noreturn in IR

kant2002:
Since codegen of inline functions was not working in #705, then instead of just removing specifiers, I put them into IR, and let somebody utilize it if needed.


ForNeVeR (commented):
Yup, a good idea.

PR #711, opened 2024-11-11
StreamHandle interop

### LINKED ISSUE #681: `StdIoFunctions.Handles` is not thread-safe and uses excessive resources
Currently we use code such as this: https://github.com/ForNeVeR/Cesium/blob/d48d71e910ba319171a40e44966833bc691f1ce2/Cesium.Runtime/StdIoFunctions.cs#L532-L533

This is not thread safe and will break if several threads open files simultaneously.

Moreover, opening and closing a file will increment the `Handles` size by one and this will never decrement back.

This should be fixed — one way or another.

I propose us to migrate to some GCHandles to pin the file handle objects instead, and pass actual pointers to the app code. But we should see how to not miss the standard handles then.

evgTSV:
Closes #681

GCHandles on StreamHandles are allocated on the unmanaged heap (`AllocHGlobal`). A pointer to the GCHandle is used to control the StreamHandler. Locks are also added for thread-safe interaction.

Keeps all standard streams (stdin, stdout, stderr) in list `Handles`

### DIFF
--- a/Cesium.Runtime/StdIoFunctions.cs
+++ b/Cesium.Runtime/StdIoFunctions.cs
@@ -22,7 +22,6 @@ public unsafe static class StdIoFunctions
         public int ErrNo { get; set; }
     }
 
-    private static object _locker = new();
     internal static readonly List<StreamHandle> Handles = [];
 
     private const int StdIn = 0;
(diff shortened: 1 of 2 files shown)


Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -668,15 +667,62 @@ public static int Remove(byte* pathname)
...
+                var gchAddr = Marshal.ReadIntPtr(handle);
+                var gch = GCHandle.FromIntPtr(gchAddr);
+
+                return gch.Target as StreamHandle;
+            }
+        }
+        catch (AccessViolationException)
+        {
+            return null;
+        }
```
ForNeVeR, line 691: There's no point in trying to catch this. The program should fail if an exception is thrown here.

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -668,15 +667,62 @@ public static int Remove(byte* pathname)
...
-        var handleIndex = (int)(IntPtr)stream;
-        var result = Handles.ElementAtOrDefault(handleIndex);
-        return result;
+        lock (_locker)
+        {
+            var gch = GCHandle.Alloc(stream);
+            var handel = GCHandle.ToIntPtr(gch);
+
+            var ptr = Marshal.AllocHGlobal(sizeof(IntPtr));
+            Marshal.WriteIntPtr(ptr, handel);
```
ForNeVeR, line 702:
```suggestion
            var handle = GCHandle.ToIntPtr(gch);

            var ptr = Marshal.AllocHGlobal(sizeof(IntPtr));
            Marshal.WriteIntPtr(ptr, handle);
```

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -668,15 +667,62 @@ public static int Remove(byte* pathname)
...
+            return ptr;
+        }
+    }
+
+    internal static bool RemoveStream(IntPtr handle)
```
ForNeVeR, line 707: Let's call this `FreeStream`?

ForNeVeR (changes requested):
Thanks! I have some minor comments. Generally this LGTM.

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -668,15 +667,62 @@ public static int Remove(byte* pathname)
         return 0;
     }
 
-    private static unsafe void* GetLastStream()
+    internal static StreamHandle? GetStream(IntPtr handle)
```
kant2002, line 670: Can we accept `void* filePtr` parameter here and perform cast inside? I think otherwise it's unnescessary noise to convert everywhere.

evgTSV:
I have included your suggestions

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -668,15 +667,66 @@ public static int Remove(byte* pathname)
...
+                var gchAddr = Marshal.ReadIntPtr(handle);
+                var gch = GCHandle.FromIntPtr(gchAddr);
+
+                return gch.Target as StreamHandle;
+            }
+        }
+        catch (InvalidOperationException)
+        {
+            return null;
+        }
```
ForNeVeR, line 693: When does this get thrown? Looks suspicious. Maybe we shouldn't handle any exception here?

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -668,15 +667,66 @@ public static int Remove(byte* pathname)
...
+
+        try
+        {
+            lock (_locker)
+            {
+                var gchAddr = Marshal.ReadIntPtr(handle);
+                var gch = GCHandle.FromIntPtr(gchAddr);
+
+                return gch.Target as StreamHandle;
+            }
```
ForNeVeR, line 688: Hm, now wait, why we need this `_locker` at all? It doesn't seem to defend any collection manipulation. Let's maybe get rid of the `_locker`?

ForNeVeR (changes requested):
Thanks! I believe we are getting closer to merging this.

Cesium.Runtime/StdIoFunctions.cs
evgTSV: Usually only AV occurs there. In this case, the error can occur due to an invalid pointer argument when converting to GCHandle, eventually the Target property will throw an AV if converting fail.

ForNeVeR (approved):
Thanks!

PR #716, opened 2024-11-14
Separate global declaration from initialization

kant2002:
That allow us have same codepath for functions initialization and for global initialization


ForNeVeR (approved):
Thanks!

PR #718, opened 2024-11-15
Initialize using address of compound initializer

kant2002:


Cesium.CodeGen/Contexts/FunctionScope.cs
```diff
@@ -138,4 +138,12 @@ public void MergeScope(BlockScope scope)
             _variableDefinition.Add(variableName, variableDefinition);
         }
     }
+
+    private int tempLocalIndex = 0;
+
+    /// <inheritdoc />
+    public string GetTmpVariable()
+    {
+        return "local_" + tempLocalIndex++;
```
ForNeVeR, line 147: I'd suggest prefixing these with `__`, to 100% avoid clash with user-defined variables. (`__`-names are reserved for us according to the language standard.)

Cesium.CodeGen/Ir/Expressions/CompoundObjectFieldInitializer.cs
```diff
@@ -20,9 +20,15 @@ internal CompoundObjectFieldInitializer(AssignmentInitializer initializer)
         Designation = initializer.Designation!;
     }
 
+    internal CompoundObjectFieldInitializer(IExpression inner, Designation designation)
+    {
+        Inner = inner;
+        Designation = designation;
+    }
+
```
ForNeVeR, line 28: I'd say we better wire these constructors together via `this()`.

PR #724, opened 2024-11-20
Implement scanf

evgTSV:
Most of the features of the `scanf` function have been implemented. Added new tests for `scanf`

### DIFF
--- a/Cesium.Compiler/stdlib/stdio.h
+++ b/Cesium.Compiler/stdlib/stdio.h
@@ -73,3 +73,12 @@ void perror(const char* s);
 
 __cli_import("Cesium.Runtime.StdIoFunctions::Remove")
 int remove(const char* pathname);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::ScanF")
+int scanf(char* format, ...);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::FScanF")
+int fscanf(FILE* stream, char* format, ...);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::FScanF")
+int vfscanf(FILE* stream, const char* format, va_list vlist);
--- a/Cesium.Compiler/stdlib/stdio.h
+++ b/Cesium.Compiler/stdlib/stdio.h
@@ -77,6 +77,9 @@ int remove(const char* pathname);
 __cli_import("Cesium.Runtime.StdIoFunctions::ScanF")
 int scanf(char* format, ...);
 
+__cli_import("Cesium.Runtime.StdIoFunctions::ScanF")
+int vscanf(char* format, va_list vlist);
+
 __cli_import("Cesium.Runtime.StdIoFunctions::FScanF")
 int fscanf(FILE* stream, char* format, ...);
 
--- a/Cesium.IntegrationTests/stdlib/io/fscanf.c
+++ b/Cesium.IntegrationTests/stdlib/io/fscanf.c
@@ -3,10 +3,10 @@
 int main(void)
 {
     FILE* tmpf = fopen("example_unique_name", "w+");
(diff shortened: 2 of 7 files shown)


Cesium.IntegrationTests/stdlib/io/fscanf.c
```diff
@@ -0,0 +1,36 @@
...
+
+    if (result != expectedArgsConsumed) return result;
+
+    printf("Converted %d fields:\n"
+            "symbols = %s\n"
+            "str1 - str2 = %s - %s\n"
+            "i = %d\n"
+            "exp = %E\n"
+            "j = %d\n"
+            "ptr = %p\n"
```
kant2002, line 31: I think you should not validate this. You have no means to check that results is the same in regular compiler and Cesium. Unless you perform cast of specific value.

Cesium.Runtime/Cesium.Runtime.csproj
```diff
@@ -13,4 +13,8 @@
         <InternalsVisibleTo Include="Cesium.Runtime.Tests" />
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageReference Include="System.Runtime.CompilerServices.Unsafe" />
```
kant2002, line 17: That dependency seems to be used only in tests. Do you really need it?

Cesium.Compiler/stdlib/stdio.h
```diff
@@ -73,3 +73,12 @@ void perror(const char* s);
...
 int remove(const char* pathname);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::ScanF")
+int scanf(char* format, ...);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::FScanF")
+int fscanf(FILE* stream, char* format, ...);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::FScanF")
+int vfscanf(FILE* stream, const char* format, va_list vlist);
```
kant2002, line 84: This is something non-standard. Do we really needed it? I'm not against supporting it, but would like to see what application we compile and where it was needed.

kant2002: Please ignore 😄

Cesium.Compiler/stdlib/stdio.h
```diff
@@ -73,3 +73,12 @@ void perror(const char* s);
 
 __cli_import("Cesium.Runtime.StdIoFunctions::Remove")
 int remove(const char* pathname);
+
+__cli_import("Cesium.Runtime.StdIoFunctions::ScanF")
+int scanf(char* format, ...);
+
```
kant2002, line 79:
```suggestion

__cli_import("Cesium.Runtime.StdIoFunctions::ScanF")
int vscanf(char* format, va_list vlist);

```

Cesium.Runtime/Cesium.Runtime.csproj
```diff
@@ -12,5 +12,4 @@
     <ItemGroup>
         <InternalsVisibleTo Include="Cesium.Runtime.Tests" />
     </ItemGroup>
-
```
kant2002, line 15: Can we remove this change too.

Cesium.Runtime/StdIoFunctions.cs
```diff
@@ -666,6 +668,451 @@ public static int Remove(byte* pathname)
...
+                    offset++;
+
+                    break;
+
+                case "llu":
+                    var uLongLongPtr = (ulong*)((long*)varargs)[argsConsumed];
+
+                    try
+                    {
+                        var numULongLong = (ulong)ParseInteger(10);
```
kant2002, line 919: You would have problem with ulong.MaxValue if given as input.

kant2002:
Thanks

PR #727, opened 2024-11-21
Remove .h files from compilation process

kant2002:
There no need to have .h files as part of compilation,since only .c files produce something of value. .h included using preprocesor if needed.

### DIFF
--- a/Cesium.Sdk.Tests/EvaluationTests.cs
+++ b/Cesium.Sdk.Tests/EvaluationTests.cs
@@ -12,7 +12,7 @@ public class EvaluationTests(ITestOutputHelper testOutputHelper) : SdkTestBase(t
     [InlineData("SimpleCoreLibraryWithHeader")]
     public async Task Evaluation_EnableDefaultCompileItems(string projectName)
     {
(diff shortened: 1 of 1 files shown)


ForNeVeR:
@kant2002 do you still think we need this?

kant2002:
I do think we need to not feed headers into Cesium from the project. From your explanation, I think this is half backed solution, and I don't have capacity to move forward with it.

ForNeVeR:
Ah, in addition, the current approach shows the header files in the IDE's project view, which is also a plus. (Not sure they'd show up without this.)

So, even when we remove them from this, we'd need to add them as some separate item set just for that purpose.

kant2002:
I hear you. But it still think that we try to scarify compatibility for our personal convenience. Whole C world compile just by sending only C files to compiler. So we should do so too. The C compilers speed up compilation by storing state somewhere else. 

We may try teach compiler to produce some output )l(list of .H files) after compilation process and then feed this to MSbuild somehow. I really don’t know how to solve this without sacrificing speed of compilation and have mental model of C compiler.

PR #730, opened 2024-11-21
Decrease Microsoft.Build.Utilities.Core to 17.11.4

YuliiaKovalova:
This MSBuild package version comes from .net9 that can cause issues in runtime due to assemblies mismatch.


ForNeVeR:
We are not against updating to .NET 9, though already noticed some issues indeed :(

I'll investigate what's the right way of managing this.

ForNeVeR:
So far it looks @kant2002 managed to fix the problems by pinning to .NET 9: https://github.com/ForNeVeR/Cesium/commit/4aada49a04e621834c42372478512ab0488e963e

Thanks for help anyway!

PR #731, opened 2024-11-21
Supply global.json

kant2002:
This only fix Sdk tests and not Integration tests which a behave funny.

I prefer to go with #729 at least initially, and if there appetite for pursuing this, continue working on this PR.


ForNeVeR (approved):
Thanks! I made some adjustments for this to be more robust in the future.

PR #734, opened 2024-11-22
Unify handling global and local declarations

### LINKED ISSUE #720: Add support for array of strings initialization
Following declaration is not currently supported.
```c
static char *values[] = {"v1","v2","v3"};
```

kant2002:
Now GlobalVariableDefinition used only during emitting phase, to indicate that we should define global field. Fixes #720


kant2002:
Better look by individual commits. Would be more clear.

ForNeVeR (commented):
Thanks, LGTM!

PR #735, opened 2024-11-22
Add support for array with const sizes represented as enums.

kant2002:
There actual 1 meaningful changes here. Passing scope to the constant evaluator, everything else is just noise.

### DIFF
--- a/Cesium.CodeGen/Ir/Declarations/LocalDeclarationInfo.cs
+++ b/Cesium.CodeGen/Ir/Declarations/LocalDeclarationInfo.cs
@@ -1,5 +1,6 @@
 using System.Globalization;
 using Cesium.Ast;
+using Cesium.CodeGen.Contexts;
 using Cesium.CodeGen.Extensions;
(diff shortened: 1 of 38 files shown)


Cesium.CodeGen/Ir/Declarations/LocalDeclarationInfo.cs
```diff
@@ -231,7 +238,7 @@ private static (IType, string? Identifier) ProcessDirectDeclarator(IDirectDeclar
                     }
                     else
                     {
-                        var constantResult = ConstantEvaluator.TryGetConstantValue(sizeExpr.ToIntermediate(), scope: null);
+                        var constantResult = ConstantEvaluator.TryGetConstantValue(sizeExpr.ToIntermediate(scope), scope);
```
kant2002, line 241: That's the change which did the trick

ForNeVeR: Yep, I see. So much work to just provide the scope here. Thanks!

PR #744, opened 2024-12-01
Add parsing of restrict

kant2002:
I do not plan to have aliasin analysis in near future, but given the code which I compile was "verified" using other compiler, let's pretend we are fine.

Also this is allows me properly annotate standard library, currenly I strip restrict from definitions, and that would be additional work in the future.


ForNeVeR (commented):
Thanks!

PR #746, opened 2024-12-01
Add limits.h

kant2002:
Most important part of this change was adding support for numerals with typed suffixes.


ForNeVeR (commented):
Great!

PR #749, opened 2024-12-01
Add conversion to long long

kant2002:
Encounter during testing with https://github.com/csmith-project/csmith


ForNeVeR (commented):
Yeah, okay. My plans to have 128-bit `long long` are diverted for now :(

PR #751, opened 2024-12-01
Declare anonymous struct array

kant2002:


Cesium.CodeGen.Tests/verified/CodeGenArrayTests.StaticStructArrayTest.verified.txt
```diff
@@ -0,0 +1 @@
+﻿emptyString
```
ForNeVeR, line 1: Huh? What happened here? It yield no codegen results?

Cesium.CodeGen.Tests/verified/CodeGenArrayTests.StructArrayTest.verified.txt
```diff
@@ -0,0 +1,2 @@
+﻿System.Void <Module>::test()
+  IL_0000: ret
```
ForNeVeR, line 2: This one is also not very useful. I assume we'd need to upgrade these tests to handle the global variables correctly.

Cesium.CodeGen.Tests/verified/CodeGenArrayTests.StaticStructArrayTest.verified.txt
kant2002: Yes. I need to check only emitted classes. Maybe we need new type of tests.

ForNeVeR: Don't worry, I'll handle this myself.

PR #754, opened 2024-12-04
Support initializers on struct arrays

kant2002:


kant2002:
Seems to be already merged :)

PR #755, opened 2024-12-04
Improve stdlibrary with time and math functions

kant2002:
Math functions on float number implemented using their double variant in .NET Standard 2.0


ForNeVeR (commented):
Thanks!

PR #763, opened 2024-12-19
(#751) Tests: update test output, log all types in CodeGenArrayTests

ForNeVeR:


Cesium.CodeGen.Tests/verified/CodeGenArrayTests.UnSignedByteArrayTest2.verified.txt
```diff
@@ -1,37 +1,49 @@
...
+      IL_0010: ret
+
+  Type: <ConstantPool>
+  Nested types:
+    Type: <ConstantPool>/<ConstantPoolItemType1>
+    Pack: 1
+    Size: 1
+  Fields:
+    <ConstantPool>/<ConstantPoolItemType1> <ConstantPool>::ConstDataBuffer0
+      Init with (UTF-8 x 1 bytes): "�"
```
kant2002, line 49: Should we write this as escape sequences to improve validation?

ForNeVeR: Yeah, would be great. I improved it a bit but skipped these so far.

PR #782, opened 2025-02-22
NuGet Publishing Infrastructure

### LINKED ISSUE #571: Publish to NuGet
Since the SDK is (almost) ready (#80), it's time to start publishing Cesium to NuGet.

While doing that, also introduce the maintainer documentation and figure out what to do with the `CesiumCompilerPackageVersion` variable.

(See `TODO[#571]` for that one.)

Publish both SDK and the project templates.

ForNeVeR:
Closes #571.


Cesium.Sdk/Sdk/Sdk.targets
```diff
@@ -16,8 +16,8 @@ SPDX-License-Identifier: MIT
     <PropertyGroup Label="CompilerInstallationProperties" Condition=" $(SkipCesiumCompilerInstallation) != 'true' ">
         <_CesiumCompilerPackageFolderName Condition="$(CesiumCompilerPackagePath) == ''">$(CesiumCompilerPackageName.ToLower())</_CesiumCompilerPackageFolderName>
         <_CesiumCompilerPackagePath>$([System.IO.Path]::Combine($(NuGetPackageRoot), $(_CesiumCompilerPackageFolderName), $(CesiumCompilerPackageVersion)))</_CesiumCompilerPackagePath>
-        <CesiumCompilerPath Condition="$(OS) != 'Windows_NT' AND $(CesiumCompilerPath) == '' AND $(_CesiumCompilerPackagePath) != '' AND $(CesiumCompilerPackageName) != ''">$(_CesiumCompilerPackagePath)/tools/Cesium.Compiler</CesiumCompilerPath>
-        <CesiumCompilerPath Condition="$(OS) == 'Windows_NT' AND $(CesiumCompilerPath) == '' AND $(_CesiumCompilerPackagePath) != '' AND $(CesiumCompilerPackageName) != ''">$(_CesiumCompilerPackagePath)\tools\Cesium.Compiler.exe</CesiumCompilerPath>
+        <CesiumCompilerRuntime>dotnet</CesiumCompilerRuntime>
+        <CesiumCompilerPath>$(_CesiumCompilerPackagePath)\tools\Cesium.Compiler.dll</CesiumCompilerPath>
```
kant2002, line 20: I think this is raise question how to distribute our compiler, only as .NET tool? how it will work when used in the CC env var in Make files for example ?

Cesium.Compiler/Cesium.Compiler.csproj
```diff
@@ -6,14 +6,29 @@ SPDX-License-Identifier: MIT
...
+        </PackageLicenseExpression>
+        <Copyright>
+            © Microsoft Corporation. All rights reserved. <!-- Microsoft.Bcl.HashCode, Microsoft.Bcl.Memory -->
+            Copyright (c) 2005 - 2015 Giacomo Stelluti Scala &amp; Contributors; <!-- CommandLineParser -->
+            Copyright (c) 2016-2025 JetBrains s.r.o. <!-- JetBrains.Annotations -->
+            Copyright 2019 Bradley Grainger; <!-- IndexRange -->
+            Copyright © 2019, 2020 Alexandre Rabérin; <!-- QuickGraph, QuikGraph.Graphviz -->
+            Copyright (c) 2021 LanguageDev team; <!-- Yoakke.* -->
+            2024-2025 TruePath contributors &lt;https://github.com/ForNeVeR/TruePath&gt;; <!-- TruePath -->
+            $(Copyright)
```
kant2002, line 21: Should we move our copyright first, and not MS, that can give wrong impression when read not carefully.

Cesium.Sdk/Sdk/Sdk.targets
ForNeVeR: This PR currently introduces three different forms (!) of distribution the compiler.
1. As a .NET tool. This is not for the SDK purposes, but for the users, so that they can `dotnet install --global Cesium.Compiler` and then `Cesium.Compiler myfile.c`. This form of distribution could be used as `CC`.
2. As a `Cesium.Compiler.Bundle`. The SDK will use this bundle for its purposes. It is not the same as the .NET tool because there's no simple way of depending the user project on the tool — you cannot inject `<PackageReference Include="Cesium.Compiler"/>` into a `.ceproj`, NuGet won't allow you.

   So, there's a separate "compiler bundle" for the purpose of SDK use, that basically includes a compiler published as a framework-dependent executable.
3. As `Cesium.Compiler.Bundle.<platform>.zip` — a group of platform-dependent self-contained bundles, for the purpose of manual/automated download+installation by the users who don't want dependency on .NET SDK.

   These can also be used as `CC`, though I'm not sure how popular this use will be.

Cesium.Compiler/Cesium.Compiler.csproj
ForNeVeR: I sorted these in chronological order, but I guess you are right and we can put ourselves first.

Cesium.Sdk/Sdk/Sdk.targets
ForNeVeR: Needless to day, I am planning to document all that in scope of this very PR.

PR #879, opened 2025-10-16
Compare signatures from different processing units.

kant2002:
Let's have Cesium process two files.
First file will have only signatures, and as such it will only declaration and empty structure, since it's only declared.

Second file will have proper definition for function, and thus we should assume that types the same. Declaration part will have type with 0 members. Definition part will have type with all members declared.

Added ability to place multi-file tests in the `.c` folder, where all C files would be fed to Cesium

Fixed emiting structure types where declaration and definition of type split across compilation modules.

### DIFF
--- a/Cesium.CodeGen/Contexts/AssemblyContext.cs
+++ b/Cesium.CodeGen/Contexts/AssemblyContext.cs
@@ -14,7 +14,6 @@ using Cesium.Core;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Cecil.Rocks;
-using QuikGraph;
 using System.Collections;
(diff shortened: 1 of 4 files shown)


ForNeVeR:
A test wouldn't hurt! Generally I think the idea is sound.

kant2002:
I do not think we have in test infra what's required to properly test. I was thinking abotu using `testname.c/` folder as some sort of project where we can have multiple files which consist of one app. Yeah we need this in our testing framework, but I really prefer it to be separately. 

It's possible that you may combine that work with multi-arch testing, but probably it can be done separately. I prefer to work on each thing separately.

PR #940, opened 2026-02-08
Make CSharpCompilationUtil disposable

### LINKED ISSUE #492: `CSharpCompilationUtil`: make it `IDisposable`
In tests, we have a class `CSharpCompilationUtil` that is used to compile C# code.

Currently, it is a static class with a bunch of functions. The tests are expected to invoke these functions and clean up after their termination (i.e. remove the compiled assemblies and whatnot).

I don't think the tests do a good job on that, and most likely they leave a lot of garbage behind.

We should make `CSharpCompilationUtil` a normal test fixture, inject it into tests, support `IDisposable` interface and make it clean up after itself automatically.

(Or perhaps just make it `IDisposable` and make the tests to manage it, without the test fixture part.)

pkazakov-dev:
And AssemblyContext too

Closes #492.

### DIFF
--- a/Cesium.TestFramework/CSharpCompilationUtil.cs
+++ b/Cesium.TestFramework/CSharpCompilationUtil.cs
@@ -13,7 +13,6 @@ using Xunit.Abstractions;
 
 namespace Cesium.TestFramework;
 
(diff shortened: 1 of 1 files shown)


ForNeVeR (approved):
Thanks!

Issue #955, opened 2026-02-14
When running Cesium.Compiler with "--ast-dump" then output file is not generated


pkazakov-dev:
Steps to repro:
1. run Compiler against `hello.c` with "--ast-dump": `dotnet run --project Cesium.Compiler -- .\Cesium.Samples\hello.c --out out.exe --nologo --ast-dump`
2. there is no ".\out.exe" file is generated
3. run Compiler against `hello.c` without "--ast-dump": `dotnet run --project Cesium.Compiler -- .\Cesium.Samples\hello.c --out out.exe --nologo`
4. there is generated ".\out.exe" file

It seems the new output file is not produced when "--ast-dump" is used.

pkazakov-dev:
OK, that makes sense! It just wasn’t transparent to me.

PR #957, opened 2026-02-16
Document --ast-dump option for diagnostic purposes

pkazakov-dev:
Resolves https://github.com/ForNeVeR/Cesium/issues/955

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ As inputs, Cesium accepts its own internal `.obj` file format produced by the `-
   - `Windows`: doesn't get detected, so it's only possible to select manually
   - `NetModule`: is a rudiment from Cecil, not supported
 - `-c`: will produce a JSON-based object file imitation in the output file. This mode is supposed to be used when using Cesium compiler as a C compiler for an existing toolset
-- `-ast-dump`: prints the Abstract Syntax Tree (AST) of the output file. No output assembly is generated. For diagnostic use only.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,7 +144,8 @@ As inputs, Cesium accepts its own internal `.obj` file format produced by the `-
   - `Console`: gets detected from an `.exe` extension
   - `Windows`: doesn't get detected, so it's only possible to select manually
   - `NetModule`: is a rudiment from Cecil, not supported
-- `-c`: will produce a JSON-based object file imitation in the output file. This mode is supposed to be used when using Cesium compiler as a C compiler for an existing toolset.
+- `-c`: will produce a JSON-based object file imitation in the output file. This mode is supposed to be used when using Cesium compiler as a C compiler for an existing toolset
+- `-ast-dump`: prints the Abstract Syntax Tree (AST) of the output file. No output assembly is generated. For diagnostic use only.
```
ForNeVeR: It seems to be `--ast-dump`, though? Double dash?

ForNeVeR (changes requested):

README.md
pkazakov-dev: Maybe we should clarify in this same PR that when using `-c`, we use the `--out` parameter to write the “object file imitation” in JSON. In this case, no executable is generated.
Also, `--ast-dump` prints the result to the console, and no executable is generated either.

PR #964, opened 2026-03-02
fix(preprocessor): prevent crash on nested macro calls with trailing whitespace

### LINKED ISSUE #963: NEsted preprocessor calls fails
Repro code which crash the compiler.
```c
#define EMPTY() 
#define DEFER(id) id EMPTY()

#define FAIL() DEFER ( EMPTY ) ( )

FAIL()
```

sputnik-mac:
## Summary

Fixes #963

### Root Cause

When a function-like macro (e.g. `EMPTY()`) appears in a replacement list that gets re-expanded, and the expansion produces a token stream that ends with trailing whitespace after the macro name, the local `Consume()` helper in `ParseArguments` would keep looping past the end of the stream, causing an `IndexOutOfRangeException`.

**Reproduction:**
```c
#define EMPTY()
#define DEFER(id) id EMPTY()
#define FAIL() DEFER ( EMPTY ) ( )
FAIL()
```

**Crash trace:**
1. `FAIL()` expands to `DEFER ( EMPTY ) ( )`
2. Each token of this replacement is expanded individually via `ExpandMacros`
3. When `DEFER(EMPTY)` is processed, the argument `EMPTY` is expanded in its own isolated token stream `[EMPTY]`
4. The result `EMPTY EMPTY()` is then re-expanded
5. The second `EMPTY` gets its own isolated stream `[EMPTY, space, EMPTY, (, )]`  
6. Inside this stream, `ExpandMacros` tries to expand the first `EMPTY`, calls `ParseArguments`, which needs to look for `(`
7. `ParseArguments`'s local `Consume()` skips whitespace — consuming `space` then attempting another `lexer.Consume()` past the end of the stream → **crash**

### Fix

- **`MacroExpansionEngine.cs`**: Add `!lexer.IsEnd` guard to the whitespace-skipping loop in the local `Consume()` helper. Add `IsEndAt(index)` guard to the local `Peek()` helper.
- **`TransactionalLexer.cs`**: Add `IsEndAt(int offset)` method for safe bounds checking at arbitrary peek offsets.
- **`PreprocessorTests.cs`**: Add regression test `NestedMacroDeferredExpansion` that exercises the fixed code path.

### DIFF
--- a/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
+++ b/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
@@ -507,6 +507,14 @@ int main() { foo(0) return 0; }
 ");
 
     [Fact, NoVerify]
+    public Task NestedMacroDeferredExpansion() => DoPreprocess(
+@"#define EMPTY()
+#define DEFER(id) id EMPTY()
+#define FAIL() DEFER ( EMPTY ) ( )
+FAIL()
+");
+
+    [Fact, NoVerify]
     public Task IfExpressionCannotConsumeNonInteger()
     {
         return Assert.ThrowsAsync<PreprocessorException>(() => DoPreprocess(
--- a/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
+++ b/Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
@@ -506,7 +506,7 @@ int main() { char* x = foo(int x; printf(""some string"")); }
 int main() { foo(0) return 0; }
 ");
 
-    [Fact, NoVerify]
+    [Fact]
     public Task NestedMacroDeferredExpansion() => DoPreprocess(
 @"#define EMPTY()
 #define DEFER(id) id EMPTY()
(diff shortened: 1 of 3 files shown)


Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
```diff
@@ -504,6 +504,14 @@ public Task ReplaceWithHashNested() => DoTest(
     public Task ReplaceWithoutParameters() => DoTest(
 @"#define foo(x)
 int main() { foo(0) return 0; }
+");
+
+    [Fact, NoVerify]
```
kant2002: Please remove `NoVerify` on test. That's for rare occasions.

Cesium.Parser.Tests/PreprocessorTests/PreprocessorTests.cs
```diff
@@ -504,6 +504,14 @@ public Task ReplaceWithHashNested() => DoTest(
     public Task ReplaceWithoutParameters() => DoTest(
 @"#define foo(x)
 int main() { foo(0) return 0; }
+");
+
+    [Fact]
+    public Task NestedMacroDeferredExpansion() => DoPreprocess(
```
kant2002, line 510: This macro should be expanded to blank space. This is not happening right now. Handling of nested expansion should be improved.

kant2002:
superseded by #989, sorry.

Issue #680, opened 2024-10-29
Automate AST generation


ForNeVeR:
In #678, we've got a manual AST generator for verification purposes.

For now, it is fully manually written, and I believe this is far from productive.

Let's consider automating it somehow. Some basic ideas:
1. Get an AST visitor and use it instead of manual enumeration of all nodes
2. Write a source gen
3. Convert to JSON and then make some basic string or JSON tree transforms

kant2002:
I think AST nodes should have information about their range and source file of origin.

PR #965, opened 2026-03-03
Create ast dumper visitor

pkazakov-dev:
Well, here we go!

Should address ~~the first part of~~ https://github.com/ForNeVeR/Cesium/issues/680


pkazakov-dev:
> Just curious where do you see the need for another AST visitor?

You mean how base AST visitor could be reused?
Could be transformation into xml, json or something else.

PR #978, opened 2026-03-29
Implement compiler warnings

evgTSV:
Part of #958

Added all warning sets (GCC-based). Actually, even though `-Wfull` doesn't exists in major C compilers, I added it as aliase to bind all kinds of warnings.

### DIFF
--- a/Cesium.Core/StringFormatExtensions.cs
+++ b/Cesium.Core/StringFormatExtensions.cs
@@ -1,0 +1,60 @@
+// SPDX-FileCopyrightText: 2026 Cesium contributors <https://github.com/ForNeVeR/Cesium>
+//
+// SPDX-License-Identifier: MIT
+
+using System.Text;
+
+namespace Cesium.Core;
+
+public static class StringFormatExtensions
+{
+    extension(ReadOnlySpan<char> str)
+    {
+        public string FromKebabToCamel()
+        {
+            if (str.IsEmpty) return string.Empty;
+
+            var sb = new StringBuilder();
+
+            foreach (var part in str.Split('-'))
+            {
+                if (part.Start.Value == part.End.Value)
+                    continue;
+
+                sb.Append(char.ToUpper(str[part.Start]));
+
+                if (part.Start.Value != part.End.Value)
+                    sb.Append(str[(part.Start.Value + 1)..part.End]);
+            }
+
+            return sb.ToString();
+        }
+
+        public string FromCamelToKebab()
+        {
+            if (str.IsEmpty) return string.Empty;
+
+            var sb = new StringBuilder();
+
+            for (int i = 0; i < str.Length; i++)
+            {
+                if (char.IsLower(str[i]))
+                    sb.Append(str[i]);
+                else if (i == 0)
+                    sb.Append(char.ToLower(str[i]));
+                else if (char.IsDigit(str[i]) && !char.IsDigit(str[i - 1]))
+                    sb.Append('-').Append(str[i]);
+                else if (char.IsDigit(str[i]))
+                    sb.Append(str[i]);
+                else if (char.IsLower(str[i - 1]))
+                    sb.Append('-').Append(char.ToLower(str[i]));
+                else if (i + 1 == str.Length || char.IsUpper(str[i + 1]))
+                    sb.Append(char.ToLower(str[i]));
+                else
+                    sb.Append('-').Append(char.ToLower(str[i]));
+            }
+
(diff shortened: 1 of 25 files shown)


Cesium.Core/Warnings/IWarningProcessor.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Cesium.Core.Warnings;
 
-public interface IWarningProcessor
+public interface IWarningProcessor<in T> where T : DiagnosticWarning
 {
-    public void EmitWarning(PreprocessorWarning warning);
+    public void EmitWarning(T warning);
 }
```
ForNeVeR, line 10: Are you sure we need a generic here? Perhaps just a `DiagnosticWarning` will be enough? They aren't treated differently anyway, are they?

Cesium.Core/Warnings/CompilerWarning.cs
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2026 Cesium contributors <https://github.com/ForNeVeR/Cesium>
+//
+// SPDX-License-Identifier: MIT
+
+namespace Cesium.Core.Warnings;
+
+public record CompilerWarning(string Message, WarningsSet Set)
+    : DiagnosticWarning(new SourceLocationInfo(string.Empty, 0, 0), Message);
```
ForNeVeR, line 8: _(Note)_ Honestly, I am not convinced that we need inheritance here. But perhaps it's okay for now, and we can decide later when we'll implement positional info for compiler warnings.

Cesium.Core/StringFormatExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+using System.Text;
+
+namespace Cesium.Core;
+
+public static class StringFormatExtensions
+{
+    extension(ReadOnlySpan<char> str)
```
ForNeVeR, line 11: Please add some tests for this stuff.

Cesium.Core/Warnings/WarningsSet.cs
```diff
@@ -0,0 +1,43 @@
+// SPDX-FileCopyrightText: 2026 Cesium contributors <https://github.com/ForNeVeR/Cesium>
+//
+// SPDX-License-Identifier: MIT
+
+namespace Cesium.Core.Warnings;
+
+[Flags]
+public enum WarningsSet : uint
+{
+    None                    = 0,
```
ForNeVeR, line 10: _(Note)_ I suppose we'll need a follow-up issues to actually implement these groups? :)

ForNeVeR (changes requested):

Cesium.Core/Warnings/IWarningProcessor.cs
evgTSV: Absolutely, we have to use a generic in order to handle different types of warnings with different metadata. Since the CompilerWarning record has `Set` field (`Diagnostic` dosen't have), we should keep it as generic.

Cesium.Core/Warnings/CompilerWarning.cs
evgTSV: I'll mention it in a follow-up issue.

Cesium.Core/Warnings/WarningsSet.cs
evgTSV: Sure

Cesium.Core/StringFormatExtensions.cs
evgTSV: Okay! I will

Cesium.Core/Warnings/IWarningProcessor.cs
kant2002: Can we have non generic interface with emitting all possible warning as methods

evgTSV: If we had a large interface with methods for all types of warings, we would constantly edit interface and its implementations on each new warning type. Or did you mean another direction (e.g have diff implementaions for file and console processing separately)? Nevertheless, I suppose that approach via generic will be better.

Cesium.Core/Warnings/WarningsSet.cs
evgTSV: https://github.com/ForNeVeR/Cesium/issues/988

ForNeVeR (approved):
Thanks!

PR #979, opened 2026-03-30
Fix `BinaryOperatorExpression`'s handling of pointer arithmetic and make `Lower()` idempotent

### LINKED ISSUE #516: `BinaryOperatorExpression`'s try to handle pointer arithmetic is problematic
In `BinaryOperatorExpression::LowerPointerArithmetics`, we are trying to essentially multiply the pointer operands by pointer size. Note that the expressions still keep the same pointer type.

This means that:
1. Initial arithmetics that's getting into the `BinaryOperatorExpression` is essentially incorrect, or at least doesn't correspond to the internal model of how the calculation is represented.
2. `Lower().Lower()` called twice on the same node would give incorrect results, since it'd lower the pointer arithmetics twice.

For example, let's consider that we are lowering the expression of `(int*)0 + 4`. In C's pointer arithmetic, this means that the result is `0 + 4 * sizeof(int*)`.

`BinaryOperatorExpression::LowerPointerArithmetics` essentially converts this to `(int*)0 + sizeof(int*) * 4`, which looks correct w.r.t. runtime code gen (.NET runtime doesn't do any pointer arithmetics under the cover), but doesn't change the types of the arguments, which is not how it should work.

Please see the code marked by number `xxx` in the compiler code, and fix that, either by introducing some new model, or removing this lowering somewhere else.

ChristosMaragkos:
Closes #516 

Previously, pointer math was lowered by multiplying the offset, but the operands retained their `PointerType`. This meant that if `Lower()` was called recursively on the same node, the compiler would identify the pointer type again and multiply the offset by `sizeof(T)` a second time, leading to corrupted memory addresses.

This PR addresses these issues by refactoring `BinaryOperatorExpression::LowerPointerArithmetics` to use an explicit calculation model based on type matching. More specifically, pointer arithmetic is now lowered as such:

1. Decay: Arrays are decayed to pointers immediately.
2. Cast to `NativeInt`: The pointer operand is cast to CTypeSystem.NativeInt.
3. Type-based scaling: The offset is multiplied (or divided for pointer subtraction) by the `SizeOfOperatorExpression` of the base type.
4. `Lower()` idempotency: Because the arithmetic now happens between a `NativeInt` and an integer, subsequent calls to `Lower()` do not trigger pointer-scaling logic again
5. Cast to original type: The final result is cast back to the original `PointerType` to maintain type safety in the AST.

In addition to these changes, I have also:
- Added and tested`Cesium.IntegrationTests/pointers/pointers_double_lowering.c` to veify that `Lower` is indeed idempotent
- Verified the IL output matches the expected lowering by checking the relevant tests

### DIFF
--- a/Cesium.IntegrationTests/pointers/pointers_double_lowering.c
+++ b/Cesium.IntegrationTests/pointers/pointers_double_lowering.c
@@ -1,0 +1,14 @@
+int main() {
+    int arr[5] = {10, 20, 30, 40, 50};
+    int *p = &arr[0];
+
+    int *target = (p + 2);
+
+    // If double-lowering occurred, target would point
+    // way past the array (p + 2 * 4 * 4)
+    if (*target == 30) {
+        return 42;
+    }
+
+    return 1;
+}


kant2002:
Thanks!

PR #981, opened 2026-03-30
Port PowerShell scripts to bash

ChristosMaragkos:
I took the liberty of porting existing powershell scripts to bash, since I found myself constantly translating anyway.

### DIFF
--- a/scripts/bash/approve_all.sh
+++ b/scripts/bash/approve_all.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+
+SOLUTION_ROOT="$SCRIPT_DIR/.."
+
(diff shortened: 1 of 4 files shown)


kant2002:
Thanks

ForNeVeR:
I am sorry but after looking at those, I have to roll this back. Current concerns:
1. I am not familiar with Bash and don't wish to support the scripts on my own. This is duplicated work for any script changes in the future.
2. `get_version.sh` is not needed at all (only used by CI, not developers)
3. `approve_all.sh` is broken (invalid syntax if you run it) — and is trivial to replicate with just a `cp` command when needed
4. `test_encoding.sh` you probably don't even need, especially if you operate in Unix environment. It also seems incorrect (case-sensitive file extension?)
5. `update_version.sh` is only needed for maintainers anyway — in addition, you'd need to add handling of `update_version.sh` into `Update-Version.ps1` and vise versa. Moreover, it's not semantically equivalent (won't throw errors if stuff was not found?)

Why did you need to "constantly translate" these?

PR #1005, opened 2026-08-14
Fix build on main

kant2002:


global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "10.0.302",
-    "rollForward": "latestFeature",
+    "version": "10.0.204",
+    "rollForward": "disable",
```
ForNeVeR, line 4: This will be very problematic, could you please explain why we need to forbid roll forward?

kant2002: Issue with build appearing on 10.0.400 so it's was faster fix. Maybe I should leave 302, but I did not have it on the laptop, so I choose what unblock build.

ForNeVeR: Automation will bump this to 400 anyway. Disabling roll forward is not a great idea anyway IMO — will cause problems for basically everyone. We need to handle this more carefully.

Issue #343, opened 2022-10-23
Allow storage classes other than static


ForNeVeR:
C standard (**6.7.1 Storage-class specifiers**) defines the following storage classes:
- `typedef` (not a real storage class)
- `extern`
- `static`
- `_Thread_local`
- `auto`
- `register`

Only one storage class per declaration is allowed (with an exception of `_Thread_local` with either `static` or `extern`).

We should allow these storage classes in the `Cesium.CodeGen.Ir.Declarations.ScopedIdentifierDeclaration`.

When implementing this, find the number `xxx` in the sources.

This issue doesn't include full processing of these storage classes, but only allowing them in the IR classes and proper parsing.

Ashish-Rawal-047:
Hi! I'd like to work on this issue.

Could you please assign it to me? Thanks!

kant2002:
It's better ask for forgiveness then for permissions. Go for PR if you can

kant2002:
Hold on. This is likely would be complicated. Take something which you can do and what did not require lot of code.

Ashish-Rawal-047:
Thanks for the suggestion! I'd still like to give this one a try—I understand it may be more involved, and I'm happy to work through it and submit a PR for review.

kant2002:
Just want you to know that we are fine with partial work

PR #1008, opened 2026-08-22
Support additional C storage class specifiers

Ashish-Rawal-047:
## Summary

This PR adds support for additional C storage-class specifiers across parsing and IR lowering.

### Changes

* Add parser support for `auto`, `register`, and `_Thread_local`
* Extend `StorageClass` and `ExtractStorageClass()` to recognize the new specifiers
* Implement end-to-end support for `auto` and `register`
* Treat `register` as automatic storage during code generation
* Report `WipException(343)` for unsupported `_Thread_local` code generation instead of internal exceptions

### Validation

* `dotnet build Cesium.sln` ✅
* `dotnet test Cesium.CodeGen.Tests` (393 tests, 0 failures) ✅

**References:** #211, #343


kant2002:
Please implement end to end support. There no point to have only parser. The real work in the implementation. 

For example register keyword likely require extensive discussion.

Ashish-Rawal-047:
Thanks for the feedback! I've updated the PR with an end-to-end implementation.

The changes now include semantic handling for `register` (including the address-of constraint), thread-local storage support through the code generation pipeline, and the corresponding tests. I'd appreciate another review.

Cesium.CodeGen.Tests/CodeGenPrimitiveTypeTests.cs
kant2002, line 31: Rebase over latest main. This should be gone

Cesium.CodeGen/Contexts/BlockScope.cs
```diff
@@ -54,12 +54,21 @@ public void AddVariable(StorageClass storageClass, string identifier, IType vari
...
                 ((IDeclarationScope) Parent).AddVariable(storageClass, identifier, variable, constant);
                 break;
+            case StorageClass.ThreadLocal:
+                // C11 §6.7.1p3: _Thread_local at block scope must be combined with static or extern.
+                throw new CompilationException(
+                    $"'_Thread_local' on block-scope variable '{identifier}' must be combined with 'static' or 'extern' (C11 §6.7.1).");
             default:
                 throw new ArgumentOutOfRangeException(nameof(storageClass), storageClass, null);
         }
+
```
kant2002, line 71: Don't insert unnecessary empty lines.

Cesium.IntegrationTests/thread_local_storage.c
```diff
@@ -0,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Cesium contributors <https://github.com/ForNeVeR/Cesium>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/* _Thread_local at file scope: each thread gets its own copy of the variable. */
+
+_Thread_local int tl_counter = 0;
```
kant2002, line 9: We need to test that this is somehow works as expected in the different threads. That probably require creating support for at least parts of `<threads.h>` which should be separate work, since it require quite a bit of testing, even if implementation should be simple.

Cesium.Parser/CParser.cs
```diff
@@ -501,11 +501,10 @@ private static InitDeclarator MakeInitDeclarator(Declarator declarator, (IToken
...
     [Rule("storage_class_specifier: 'extern'")]
     [Rule("storage_class_specifier: 'static'")]
-    // TODO[#211]:
-    // storage-class-specifier:
-    //     _Thread_local
-    //     auto
-    //     register
+    [Rule("storage_class_specifier: '_Thread_local'")]
+    [Rule("storage_class_specifier: 'auto'")]
+    [Rule("storage_class_specifier: 'register'")]
```
kant2002, line 506: For each storage modifier I would like to see separate PR, since I want to validate independently what's going on. For example we need separate tests for auto for register and for thread_local. Also your impleemntation only use old pre-C23 identifier, but we target C23 so we should include new one.

I would put register on hold, since we need to properly implement auto and then maybe we can go with almost same implementation as auto.

ForNeVeR (changes requested):
For the record, I think we can mostly ignore `register`, just forbid certain interactions with it (please somebody double-check my reading of the standard).

I agree that it'd be better to split this into several smaller pull requests, though. Could you please do that?

Issue #372, opened 2023-04-30
Function and type namespaces for outside usage


ForNeVeR:
Most of the current major C implementations only deal with external function visibility and not namespaces. In this issue, by _external visibility_ I mean visibility to the outside world, i.e. the possibility to call a function from an external program (by linking a C object or a DLL).

There are two approaches:
1. Make everything "public" (i.e. callable from the outside) by default. This strategy is used by GCC and other Linux-originating compilers.
2. Make everything "private" by default, introduce a special way (i.e. an `__attribute__` or `__declspec(dllexport)`) to designate the public symbols.

In both cases, it's possible to change the default behavior via a toolset setting, and I think it's also possible to explicitly set the list of public symbols via some additional files that get processed at the linking stage.

For ease of external use, a header file may be provided by a C program that contains a set of definitions in which the functions are marked by a dual attribute (i.e. `__declspec(dllimport)`).

Cesium is different from classic C implementations in several aspects.

1. We have no separate linking stage, as there is no analog of object files in .NET.
2. Among the _functions_, we also have to export _types_ from the resulting assembly for it to be useful to other .NET languages.
3. We provide no header files for the external programs to use Cesium programs, but rely on public .NET metadata instead.

I consider usability of Cesium-compiled programs by external .NET code as an issue of critical importance, and I want as much as possible to work flawlessly by default for major use-cases of Cesium.

But that means we'll have to deal with much more problems than C normally do.

## Different type definitions across translation units
The main problem is that a C program may be composed of several translation units, each of which may provide different definitions for symbols with the same name. And while this shouldn't be an issue for public functions (since such programs would be considered broken at the linking stage), this may be an issue for type definitions. Consider this:
```c
// a.c
struct foo { int x; };
void perform1(struct foo*) { }
```
```c
// b.c
struct foo { double x; };
void perform1(struct double*) { }
```

It is possible to compile a valid C program from these definitions, but it is pretty hard to express this program in .NET metadata.

## C namespaces
C17 doesn't formally provide any "namespaces", but the type symbols in the language are generally divided into two categories: `typedef`'d names and type names. These two categories form two distinct "namespaces" (not in terms of C but in terms or programming language design): in the scope of a translation unit, you can only have one entity per name per namespace.

E.g. this is valid:
```
struct s { int x; };
typedef struct { int y; } s;

struct s a; // has member x
s a; // has member y
```

While this is not:
```
union s { int x; };
[...]

PR #1015, opened 2026-09-06
Support exporting enum types via assembly metadata

### LINKED ISSUE #1012: Enum types should be exported via the assembly metadata
Imagine the following C program:
```c
typedef enum Side {
    Side_L,
    Side_R
} Side;
```

I expect `Side_L` and `Side_R` to be available for .NET side: either `Side` type should be exported, or at least constants `Side_L` and `Side_R` should exist.

I propose: let's start by exposing this as a real `enum` type.

Note that the scope of this issue is to export the type purely for .NET interop purpose: this doesn't mean that the C functions accepting this enum's values have to be converted as well. It would be nice, surely, but I'm not sure what the consequences for the function pointer compatibility this might be. So, for now, let's just update the CLR metadata, and not use the values of this enum anywhere in the Cesium-generated IL code.

Dread63:
Closes #1012 

Exports C `enum` declarations as CLR `enum` types. This makes `enum` member types visible in the .NET intermediate code (Side.Side_L for example) as suggested in the issue.

These changes are meta-data only, so `enum` values continue to be treated as Int32 values in generated IL code.

Supporting changes required to make this work:
- `CesiumMetadataImporterProvider`: Added `System.Enum` to the reference assembly list
- `BlockItemEmitting`: `enum` tag/typedef declarations were not sent to `GenerateType` (only structs were). Added calls for `EnumType`

Test cases:
- `EnumTypeTest`/`EnumTypedefTest`: Validate tag and typedef metadata consistency
- `EnumInitializerTest`: Validate auto increment correctly continues after writing metadata for an explicit initializer (`Enum Colors { Red, Green = 5, Blue };` Where blue then has a corresponding value of 6)
- Updated `CodeGenArrayTests.EnumArraySizeTest`: Now declares named `enum` used to size an array and emits correct metadata
- Updated `CodeGenTestBase.cs`: Extended `DumpTypes` to print `field.constant` as an extra layer of verification

Verified via `dotnet nuke TestAll`

### DIFF
--- a/Cesium.CodeGen.Tests/CodeGenEnumTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenEnumTests.cs
@@ -17,6 +17,17 @@ public class CodeGenEnumTests : CodeGenTestBase
         return VerifyMethods(moduleType);
     }
 
+    [MustUseReturnValue]
+    private static Task DoTypeTest(string source)
+    {
+        var assembly = GenerateAssembly(default, source);
+        return VerifyTypes(assembly);
+    }
+
+    [Fact]
+    public Task EnumTypeTest() => DoTypeTest(@"
+typedef enum Side { Side_L, Side_R } Side;");
+
     [Fact]
     public Task EnumDeclarationTest() => DoTest(@"
 enum Colour { Red, Green, Blue };
--- a/Cesium.CodeGen.Tests/CodeGenEnumTests.cs
+++ b/Cesium.CodeGen.Tests/CodeGenEnumTests.cs
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2025 Cesium contributors <https://github.com/ForNeVeR/Cesium>
+// SPDX-FileCopyrightText: 2025-2026 Cesium contributors <https://github.com/ForNeVeR/Cesium>
 //
 // SPDX-License-Identifier: MIT
 
@@ -17,7 +17,6 @@ public class CodeGenEnumTests : CodeGenTestBase
         return VerifyMethods(moduleType);
     }
 
-    // Currently failing tests using DoTypeTest is marked as "succeeded" so its obvious when proper implementation works in the future
     [MustUseReturnValue]
     private static Task DoTypeTest(string source)
     {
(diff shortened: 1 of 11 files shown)


ForNeVeR:
One of the integration tests has failed — `sizeof.portable.c`:
```
[ERR] Unhandled exception. System.BadImageFormatException: Duplicate type with name 'foo' in assembly 'out_cs, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

What is the current naming scheme for enums? I am starting to think that perhaps we'd need to prefix them somehow.

Not sure if this is a conflict in standard C or not:
```c
enum E { A, B, C };
struct E { int x; };
```

If it is _not_ a conflict in standard — this means we'd need to have some naming convention preventing this; I'd propose having `_Enum_` prefix added to metadata names of such types. Quite ugly (I'll admit!), but guaranteed by the standard to not be occupied by any user-defined type or identifier.

Dread63:
> If it is _not_ a conflict in standard — this means we'd need to have some naming convention preventing this; I'd propose having `_Enum_` prefix added to metadata names of such types. Quite ugly (I'll admit!), but guaranteed by the standard to not be occupied by any user-defined type or identifier.

Happy to implement this change. What do you think of only applying the _Enum_ prefix if the enum's name collides with that of another type? E.g. 'Enum foo{..}' and 'Struct foo {...}' but otherwise leaving the prefix out to maintain the cleaner 'Side.Side_L' view requested in the original issue?

kant2002:
Behavior should be consistent. Always prefix, or no prefix. If struct and enum namespaces unique per spec, that will require separate PR with tests, since it's unrelated to enum interop.

ForNeVeR:
Yeah, let's have the prefix always.

I wouldn't like the public API shape to depend on presence of unrelated types in the assembly.

Dread63:
Will do! I'll fix and push today!

ForNeVeR:
JFYI, tests still experience problem in the same integration test file.

Interestinhgly, it seems like your changes are applied, but we still see a duplicate type!

```
Unhandled Exception: System.BadImageFormatException: Duplicate type with name '_Enum_foo' in assembly 'out_cs, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

Would you please investigate it a bit more? What's the source of duplication? Depending on that, we can come up with a plan.

Dread63:
Will do, I'll investigate and see if i can figure out what's going on!

Dread63:
@ForNeVeR looks the issue was an oversight by me. At some point this check was implemented for struct types as well. Inside Cesium.CodeGen/Contexts/TranslationUnitContext.cs the resolveType function handled forward declarations for structs but not enums. Looks like the updated branch passes all checks. Let me know if anything needs to change!

ForNeVeR (approved):
Thanks! I have some further ideas on how to make naming less ugly — wrote them in #372. For now, this will work fine.

